PR #354, opened 2015-01-05
pyfaf: fix utf-8 handling in filename field when doing a bthash

mbrysa:


sorki:
+1. Fix quotes and push it

PR #355, opened 2015-01-05
Webfaf2 improve filters

mbrysa:


sorki:
Newly added stuff is missing in makefile & spec

mbrysa:
I amended the two Makefiles. Changes to spec are not necessary, as we use `*.css` and `*.js`.

sorki:
What's the reason behind changing Component multiselect? For me I doesn't look like an improvement - actually quite the opposite as it looks inconsistent compared to surrounding fields, there's no way to go through a list of components, I allows arbitrary values and usage is rather cumbersome compared to previous version.

mbrysa:
1. It takes about 5 seconds to open the dropdown with current Fedora components on not-so-old PCs.
2. This way people can paste in a comma-separated text list of components that they already have in their BZ query.

I'm of course open to suggestions on how to do in a more UX consistent way.

PR #375, opened 2015-01-26
Webfaf2 tweaks3

mbrysa:


sorki:
Please add explanation of 'everyone_is_maintainer' option to config as it's not very descriptive by itself. Otherwise fine.

PR #388, opened 2015-02-12
PostgreSQL by default

sorki:
- used for testing
- required in spec

- python-testing.postgresql now being pushed to fedora and el7
- el6 package https://copr.fedoraproject.org/coprs/rmarko/python-testing.postgresql/ (requires https://copr.fedoraproject.org/coprs/rmarko/python-psycopg2/ )


sorki:
Obsoleted by semnatic_pg branch.

PR #390, opened 2015-02-19
Semantic versioning

### LINKED ISSUE #334: Filter problems by NEVR
Filter problems by specifying a range of versions.

sorki:
storage: add Build.semver (semantic version field)
    
    Requirement for comparing versions on database level.
    
    Uses pg-semver (https://github.com/theory/pg-semver/)
    PostgreSQL extension implementing Semantic version
    specification (http://semver.org/).
    
    Closes #334.
 
Also contains commits from PR #388 as it depends on it (these commits will disappear as it's merged).

How to test:

1) install packages

2) as admin user of postgresql run 

    CREATE EXTENSION semver;

this can't be done via migration as only admin is permitted to install extensions.

3) run migration
4) verify via 

    SELECT * FROM builds ORDER by base_package_name, semver DESC LIMIT 1000;

or via `faf shell`

    session.query(Build.semver).order_by(Build.semver).limit(100).all()


src/pyfaf/storage/events.py
```diff
@@ -36,3 +37,16 @@ def before_delete(mapper, connection, target):
...
+@event.listens_for(Session, "before_flush")
+def store_semantic_version_for_build(session, flush_context, instances):
+    """
+    Store semnatic version (Build.semver) converted from text version
+    (Build.version)
+    """
+
+    objects = session.new.union(session.dirty)
+    for obj in [c for c in objects if isinstance(c, Build)]:
+        if isinstance(obj, Build):
```
mbrysa: One of these `isinstance` checks seems redundant. I'd even say the whole generator is. 
Also, do we want to do it like this using an event? Would a property setter on the Build.version be perhaps better?

mbrysa: I guess the property setter would mess with the ORM too much.
What about http://docs.sqlalchemy.org/en/rel_0_7/orm/events.html#sqlalchemy.orm.events.AttributeEvents.set ? This event seems a little more fine-grained so it might have less performance impact.

mbrysa:
Successfully builds for me on RHEL 6 metal and COPR and F21!
I would just add the `CREATE EXTENSTION` stuff into `abrt-server-install`.

sorki:
Sounds good. Will add it to both abrt-server-install and ansible recipes.

sorki:
Indeed. Good catch.

sorki:
Fixed.

mbrysa:
I tested the migration and it seems to work. (There's a separate issue with an older migration that makes it only work with alembic>=0.5.0, which is not in EPEL)

However from a usefulness point of view I wonder whether also not use semver for Build.release. In the Enterprise Linux world, many of the packages keep their versions for eternity and only change release numbers (kernel is the best example). If we want maintainers to be able to filter out older package releases, they need to specify release number, because package version is the same.
We would then be able to query like `(Build.version == '3.10.0' and Build.release > '123.el7') or Build.version > '3.10.0'` to get builds never than 3.10.0-123.el7.

mbrysa:
Thanks a lot!

PR #400, opened 2015-03-04
Problem version filter

sorki:


src/webfaf2/problems.py
```diff
@@ -79,7 +82,6 @@ def query_problems(db, hist_table, hist_column,
     if associate_id:
         assoc_query = (
             db.session.query(ProblemComponent.problem_id.label('problem_id'))
-            .distinct(ProblemComponent.problem_id)
```
mbrysa: This for some reason reverts my earlier commit https://github.com/abrt/faf/commit/8a28a3a07c06daa7a11bcacff3a90f9c7d279100 .

src/webfaf2/templates/problems/list.html
```diff
@@ -32,6 +32,13 @@
                 {{ render_field(filter_form.source_file_names, placeholder="Source files") }}
                 <p class="text-muted">Comma acts as OR, you can use % as a wildcard, at least one backtrace where all three fields are satisfied must exist to show the problem</p>
               </div>
+              <div>
+                {{ render_field(filter_form.since_version, placeholder="Since version") }}
+                {{ render_field(filter_form.since_release, placeholder="Since release") }}
+                {{ render_field(filter_form.to_version, placeholder="To version") }}
+                {{ render_field(filter_form.to_release, placeholder="To release") }}
+                <p class="text-muted">Version range</p>
```
mbrysa: Please also add the new fields to line 17 to make the advanced filters stay open.

src/webfaf2/problems.py
```diff
@@ -151,6 +153,30 @@ def query_problems(db, hist_table, hist_column,
...
+        final_query = final_query.filter(Problem.id == ver_sq.c.problem_id)
+
+    if to_version:
+        ver_sq = version_query.filter(Build.semver <= to_version).subquery()
+        final_query = final_query.filter(Problem.id == ver_sq.c.problem_id)
+
+    if to_release:
+        ver_sq = version_query.filter(Build.semrel <= to_release).subquery()
+        final_query = final_query.filter(Problem.id == ver_sq.c.problem_id)
+
```
mbrysa: This requires both the version AND release to be higher than specified. If we have two problems with package versions 3.10.0-10 and 3.11.0-5, then searching this way for newer than 3.10.0-9 doesn't yield the 3.11.0-5, because it's release is less than 9, even though the more significant version is higher. The correct way was outlined here https://github.com/abrt/faf/pull/390#issuecomment-76385903 .

Also this creates a whole new subquery for every parameter used, which seems costly. I'd rather put all the filters into one subquery and apply it only once to the `final_query`, as is the case with the `function_names` etc. filter just above.

sorki:
Fixed issues found by review.

mbrysa:
Thank you!

PR #404, opened 2015-03-10
c2p: do not die when eu-unstrip returns non-printable characters

mtoman:
Fixes RHBZ#1199509

Signed-off-by: Michal Toman <michal.toman@gmail.com>


mbrysa:
Thank you!

PR #407, opened 2015-03-18
pyfaf: add reposync-light action

mbrysa:
Allows syncing of builds from a repo without downloading the
packages. Useful for OpenShift or Docker instances where storage
is limited and retracing (where downloaded packages are necessary)
is not being done anyway.

### DIFF
--- a/faf.spec.in
+++ b/faf.spec.in
@@ -316,6 +316,13 @@ Requires: %{name} = %{faf_version}
 %description action-repo
 A plugin for %{name} implementing repoadd, repolist and reposync actions
 
+%package action-reposync-light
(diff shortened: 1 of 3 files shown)


mtoman:
Wouldn't a switch to the existing `reposync` be a better option? Formerly I have also been thinking of a switch that would download the RPM, save dependencies into DB and then delete the RPM.

mbrysa:
Hmm, that makes sense.
`--no-download-rpm` to do what this does now and
`--no-store-rpm` to do what you suggest.

mbrysa:
Superseded by #408 .

PR #409, opened 2015-03-23
CentOS MantisBT support

mbrysa:


src/pyfaf/bugtrackers/mantisbt.py
```diff
@@ -0,0 +1,246 @@
...
+# faf is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with faf.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import
+#from __future__ import unicode_literals
```
sorki: leftover

src/pyfaf/bugtrackers/mantisbt.py
```diff
@@ -0,0 +1,246 @@
...
+    "open": None,
+    "fixed": "UPSTREAM",
+    "reopened": None,
+    "unable to reproduce": "WORKSFORME",
+    "not fixable": "CANTFIX",
+    "duplicate": "DUPLICATE",
+    "not a bug": "NOTABUG",
+    "suspended": "INSUFFICIENT_DATA",
+    "won't fix": "WONTFIX"
+}
```
sorki: Why is this necessary? If you create separate table for Mantis bugs why there's no 1:1 mapping of states and resolutions?

src/pyfaf/actions/attach_centos_bugs.py
```diff
@@ -0,0 +1,71 @@
...
+                           get_reportmantis,
+                           get_osrelease)
+from pyfaf.storage import ReportMantis
+
+
+class AttachCentosBugs(Action):
+    """
+    Looks up all bugs created using ABRT on CentOS MantisBT and attaches them
+    to reports using the bthash from URL. This is a workaround until proper
+    attaching is supported by the ABRT client.
```
sorki: I'm wondering why is this (reporting to mantis) even deployed without "proper attaching". Agile strike?

src/pyfaf/storage/mantisbt.py
```diff
@@ -0,0 +1,76 @@
...
+    status = Column(Enum(*BUG_STATES, name="mantisbug_status"), nullable=False)
+    resolution = Column(Enum(*BUG_RESOLUTIONS, name="mantisbug_resolution"), nullable=True)
+    duplicate = Column(Integer, ForeignKey("{0}.id".format(__tablename__)), nullable=True, index=True)
+    creation_time = Column(DateTime, nullable=False)
+    last_change_time = Column(DateTime, nullable=False)
+    tracker_id = Column(Integer, ForeignKey("{0}.id".format(Bugtracker.__tablename__)), nullable=False, index=True)
+    opsysrelease_id = Column(Integer, ForeignKey("{0}.id".format(OpSysRelease.__tablename__)), nullable=False, index=True)
+    component_id = Column(Integer, ForeignKey("{0}.id".format(OpSysComponent.__tablename__)), nullable=False, index=True)
+    # whiteboard = Column(String(256), nullable=False)
+
```
sorki: This looks pretty much the same as BzBug except for creator field. Do we really want to add one more table when it's already differentiated by `tracker_id`? It would make sense if resolution and states were different but when it's only missing creator field and overrides __str__ I'm not so sure.

src/pyfaf/actions/attach_centos_bugs.py
jfilak: Exactly, why we should wait for next 6 months if we can deploy a working version of the centos reporter and see if the "proper attaching" is even necessary? If you don't like agile, stop doing agile yourself and the rest of the will follow you :-P

src/pyfaf/bugtrackers/mantisbt.py
mbrysa: There are some places where the Bugzilla status strings are being used:
https://github.com/abrt/faf/blob/master/src/webfaf2/filters.py#L87
https://github.com/abrt/faf/blob/master/src/webfaf2/templates/_helpers.html#L175
I'd say it makes sense to keep the resolutions same as well.

src/pyfaf/storage/mantisbt.py
mbrysa: Well, it's not differentiated by `tracker_id`, because `tracker_id` is not part of the primary key.
Since we're using Bugzilla IDs and Mantis IDs directly, there would be a possibility of a number conflict if they were on the same ID sequence.
We're already running two bugtrackers on one ID sequence (fedora-bugzilla and rhel-bugzilla) and we've not seen conflicts only because they by coincidence share the same Bugzilla instance.

I guess the correct thing to do here is to make `tracker_id` part of the primary key, and pull the `__str__`'s prefix from the bugtracker instance based on the tracker ID (the same way as the `url` property). Some refactoring of the `bugtrackers/mantisbt.py` plugin to take the new primary key into account would also be necessary. We could then use is as a pattern for refactoring the Bugzilla stuff the same way.

src/pyfaf/actions/attach_centos_bugs.py
sorki: It's obviously necessary if you don't want to piss off people who run mantis and put unnecessary load on their system which is EXACTLY WHY THIS PROJECT EXISTS.

It also speeds the whole reporting process but you know that very well.

But yeah, attaching is extremely complicated process, similar to testing the stuff before it's deployed. It's much easier to deliver incomplete stuff and call it a feature, eh. agile.

And I'm not doing agile for a while and won't ever do it again because it's a stupid management tool to "motivate" developers to cram more features into their status reports. And development is not only about adding features.

src/pyfaf/bugtrackers/mantisbt.py
sorki: Yeah, I've also tried to find these but did only find that one in template which is CLOSED that should be uniform across bugtrackers. The one in filter is not really important as it only maps resolutions to bootstrap label names.

So I'm not really convinced that it's worth mapping stuff to what's used in bugzilla as we loose some potentially valuable information like `feedback` status. Also how do you map back if you want to alter these bugs from the server?

src/pyfaf/actions/attach_centos_bugs.py
sorki: And if you didn't notice it also leads to unnecessary workarounds like this one that has to be done by someone else because someone didn't do his job.

src/pyfaf/storage/mantisbt.py
sorki: That makes sense, I'm not sure if it's worth doing right now - I would be happy if we just keep states and resolution as used in mantisbt so that new table actually makes sense.

I'll create an issue for this as the current implementation doesn't allow us to add another bugzilla/mantisbt different from that we already use.

src/pyfaf/actions/attach_centos_bugs.py
jfilak: ```attaching is extremely complicated process```
So you think we are idiots and we didn't implemented the attaching just for fun?

``` the stuff before it's deployed```
Why are you bringing this up? There were several test days of the CentOS integration and I don't remember you taking part of any.

```call it a feature```
We know that the integration is incomplete and we don't call it a feature. We just got a more priority task.

We do what users request and managers order.  Feel free to step up and become a leader.

sorki: > So you think we are idiots and we didn't implemented the attaching just for fun?

No I'm saying you've deployed an incomplete feature and you're doing this over and over again.

> Why are you bringing this up? There were several test days of the CentOS integration and I don't remember you taking part of any.

I prefer automation and I've invested a LOT of time into it so I don't have to do things like test days..

> We know that the integration is incomplete and we don't call it a feature. We just got a more priority task.

Like what? Adding new features?

> We do what users request and managers order. Feel free to step up and become a leader.

Maybe use your head as well from time to time.

jfilak: > No I'm saying you've deployed an incomplete feature and you're doing this over and over again.

You are right. That's my fault.

The attaching was designed:
https://github.com/abrt/libreport/issues/304
and it had to be temporarily disabled:
https://github.com/abrt/libreport/commit/f111a471853c7077127eb9b07c9adaa14c5f178c 

I've made these decisions and I took full responsibility for them.

> I prefer automation and I've invested a LOT of time into it so I don't have to do things like test days..

I really appreciate your efforts. You have done a great work! I use it a lot and I add new test cases for bugs and features. But I also appreciate things like test days.

> Like what? Adding new features?

It might appear I like new features but I'd rather stabilize the project and fix the bugs. Unfortunately, I am an loyal employee and my employers often have different orders for me. I cannot recollect any new feature I've develop where I'm also author and initiator.

> Maybe use your head as well from time to time.

I do use my head but that's obviously not enough.  Hence, I'm kindly asking you to ACK or NACK all my future pull request.

sorki:
Thanks!

PR #412, opened 2015-03-26
webfaf2: show frame addresses and build IDs

### LINKED ISSUE #389: Provide full core_backtrace
Users might be able to do their own retracing ;)

mbrysa:
For users to have at least some information about the frame in
case retracing didn't succeed.
Closes #389

### DIFF
--- a/src/webfaf2/templates/_helpers.html
+++ b/src/webfaf2/templates/_helpers.html
@@ -140,8 +140,26 @@
 {% if type != 'PYTHON' %}
   <td>{{ frame.symbolsource.path }}</td>
 {% endif %}
-<td>{{ frame.symbolsource.source_path or '-' }}</td>
-<td>{{ frame.symbolsource.line_number or '-' }}</td>
+<td>
+  {%- if frame.symbolsource.source_path %}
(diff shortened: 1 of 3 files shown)


src/webfaf2/templates/_helpers.html
```diff
@@ -140,8 +140,26 @@
...
 {% endif %}
-<td>{{ frame.symbolsource.source_path or '-' }}</td>
-<td>{{ frame.symbolsource.line_number or '-' }}</td>
+<td>
+  {%- if frame.symbolsource.source_path %}
+    <abbr title="{{ frame.symbolsource.offset|memory_address }} Build id:{{frame.symbolsource.build_id}}">{{ frame.symbolsource.source_path }}</abbr>
+  {%- elif frame.symbolsource.offset %}
+    {{ frame.symbolsource.offset|memory_address }}
+    {%- if frame.symbolsource.build_id %}
+      Build id:{{frame.symbolsource.build_id}}
```
sorki: adding <br /> before Build id improves readability a lot and results in less random line breaks.

sorki: also sed -i 's/id:{/id: {/'

PR #415, opened 2015-04-08
pyfaf: add configurable root for yum plugin

mbrysa:


sorki:
What's the use case for this? It would be nice if it is mentioned in commit message or referenced in a ticket.

mbrysa:
Reworded the commit message.

PR #423, opened 2015-04-16
pyfaf: better reuse old problems in create-problems

mbrysa:
Previously an old problem had to match in at least a half of
its reports to be reused. This methods reuses the old problem
if at least one report matches. Also properly connects the best
matches together.
This should reduce the number of empty problems left after a run
of create-problems.

### DIFF
--- a/src/pyfaf/actions/create_problems.py
+++ b/src/pyfaf/actions/create_problems.py
@@ -18,6 +18,7 @@
 
 import satyr
 from collections import defaultdict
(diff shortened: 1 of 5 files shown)


sorki:
Simple test that would cover at least part of the clustering code paths would be nice.

mbrysa:
Test for clustering added.

sorki:
Nice. Thank you.

PR #428, opened 2015-04-27
pyfaf: add retrace --max-fail-count

mbrysa:
This option allows to stop retracing a symbol after a given count of failed runs. In the production DB there are many symbols that fail over and over and eat the majority of processing time. The action should still be run occasionally without this option to to retry retracing of all symbols.

### DIFF
--- a/src/pyfaf/problemtypes/kerneloops.py
+++ b/src/pyfaf/problemtypes/kerneloops.py
@@ -514,17 +514,19 @@ class KerneloopsProblem(ProblemType):
 
         return ret_db_reports, distances
 
-    def get_ssources_for_retrace(self, db):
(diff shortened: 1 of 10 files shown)


src/pyfaf/problemtypes/kerneloops.py
```diff
@@ -633,6 +638,7 @@ def retrace(self, db, task):
                     results.reverse()
                 except FafError as ex:
                     self.log_debug("addr2line failed: {0}".format(str(ex)))
+                    db_ssource.retrace_fail_count = db_ssource.retrace_fail_count+1
```
sorki: Please use += 1 instead of these ^^, otherwise ok.

PR #431, opened 2015-04-30
Add fedmenu but display it only if configured for FAF

pypingou:
With this change, we are adding on the bottom right side of the screen
a small menu, specific to Fedora that provides a way for Fedora users
to quickly navigate through our applications.

### DIFF
--- a/src/webfaf2/templates/base.html
+++ b/src/webfaf2/templates/base.html
@@ -91,5 +91,23 @@
           </a>
         </div>
       </div>
+      {%- if config['FEDMENU_URL'] and config['FEDMENU_DATA_URL'] %}
+      <script src="{{ config['FEDMENU_URL'] }}/js/fedmenu.js"></script>
+      <script>
+        fedmenu({
+          'url': '{{ config["FEDMENU_DATA_URL"] }}',
+          'mimeType': 'application/javascript',
+          'position': 'bottom-right',
+          {%- if component %}
+          'package': '{{ component.name }}',
+          {%- elif problem %}
+            {%- if problem.unique_component_names | length == 1 %}
+            {#- unique_component_names is a set, this gets the one item #}
+            'package': '{{ problem.unique_component_names|first }}'
+            {%- endif %}
+          {%- endif %}
+        });
+     </script>
+     {%- endif %}
   </body>
 </html>


pypingou:
If you wish to test this change, you can add the following entry to the configuration file:
```
FEDMENU_URL = 'https://apps.fedoraproject.org/fedmenu'
FEDMENU_DATA_URL = 'https://apps.fedoraproject.org/js/data.js'
```
From there, the menu should become visible :)

trapas:

src/webfaf2/templates/base.html
```diff
@@ -91,5 +91,15 @@
           </a>
         </div>
       </div>
+      {% if config['FEDMENU_URL'] is defined %}
```
mbrysa: I'd prefer `{% if config['FEDMENU_URL'] and config['FEDMENU_DATA_URL'] %}` here so that `False` or `""` doesn't enable the block.
Or is there a reason for checking strictly for definition that I'm missing?

sorki:
add to whitelist

ralphbean:
I think we can get a much nicer menu if we can tell the fedmenu shim about what package is being displayed.  How about adding this?

```diff
diff --git a/src/webfaf2/templates/base.html b/src/webfaf2/templates/base.html
index d220947..11a144c 100644
--- a/src/webfaf2/templates/base.html
+++ b/src/webfaf2/templates/base.html
@@ -98,6 +98,13 @@
           'url': '{{ config["FEDMENU_DATA_URL"] }}',
           'mimeType': 'application/javascript',
           'position': 'bottom-right',
+          {% if component is defined %}
+          'package': '{{ component.name }}',
+          {% else if problem is defined %}
+            {% if problem.unique_component_names | length == 1 %}
+            'package': '{{ problem.unique_component_names[0] }}',
+            {% endif %}
+          {% endif %}
         });
      </script>
      {% endif %}
```

That should show the package sub menu on the report page (since is has a ``component`` in the jinja context) and the problem page should work too in *most* cases.  I added that check for ``length == 1`` since fedmenu only knows how to deal with one package at a time.  When there's more than one package associated with a problem.. we can just drop that part of the menu.

Any thoughts on this?

sorki:
DEBUG:   File "/builddir/build/BUILD/faf-0.12.244.gacce.dirty/src/webfaf2/templates/base.html", line 103, in template
DEBUG:     {% else if problem is defined %}
DEBUG: TemplateSyntaxError: expected token 'end of statement block', got 'if'

mbrysa:
Good idea with the component names! There were however a few problems with the commit. I'd say the final commit should look like this:
```
      {%- if config['FEDMENU_URL'] and config['FEDMENU_DATA_URL'] %}
      <script src="{{ config['FEDMENU_URL'] }}/js/fedmenu.js"></script>
      <script>
        fedmenu({
          'url': '{{ config["FEDMENU_DATA_URL"] }}',
          'mimeType': 'application/javascript',
          'position': 'bottom-right',
          {%- if component %}
          'package': '{{ component.name }}',
          {%- elif problem %}
            {%- if problem.unique_component_names | length == 1 %}
            {#- unique_component_names is a set, this gets the one item #}
            'package': '{{problem.unique_component_names|first}}'
            {%- endif %}
          {%- endif %}
        });
     </script>
     {%- endif %}
```

mbrysa:
Could you then please squash the commits into one to reduce the noise?

pypingou:
Fixed and rebased :)

mbrysa:
Thank you very much!

pypingou:
On Thu, May 07, 2015 at 12:16:03AM -0700, Marek Bryša wrote:
>    Thank you very much!

Thank you ! :)

mbrysa:
Deployed and works nice!
e.g. https://retrace.fedoraproject.org/faf/reports/167919/

pypingou:
Cool! Looks great! :)

ralphbean:
Nice :)

PR #449, opened 2015-08-05
pyfaf: pickle saved report ids only once in pull-reports

mbrysa:
The pull-reports action uses pickle to save the set of already pulled reports
to avoid pulling duplicated. Since the pickle can grow in size a lot, it can
take a couple of seconds to save to the disk. This change makes the pickle to
be saved only once after pulling all the reports instead of saving it after
every single pull. This significantly improves the speed of pulling.

### DIFF
--- a/src/pyfaf/actions/pull_reports.py
+++ b/src/pyfaf/actions/pull_reports.py
@@ -141,10 +141,8 @@ class PullReports(Action):
             return 0
 
         pulled = 0
-        i = 0
         try:
-            for report in sorted(reports):
-                i += 1
+            for i, report in enumerate(sorted(reports)):
                 self.log_debug("[{0} / {1}] Pulling {2}"
                                .format(i, len(reports), report))
                 ureport = self._get_report(report)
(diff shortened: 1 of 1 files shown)


src/pyfaf/actions/pull_reports.py
```diff
@@ -22,7 +22,7 @@
 import urllib2
 import uuid
 from pyfaf.actions import Action
-from pyfaf.common import ensure_dirs
+from pyfaf.common import ensure_dirs, FafError
```
mbrysa: This import was missing.

src/pyfaf/actions/pull_reports.py
```diff
@@ -142,25 +142,27 @@ def run(self, cmdline, db):
...
-            while os.path.isfile(filename):
-                filename = os.path.join(self.incoming_dir, uuid.uuid4().get_hex())
-
-            with open(filename, "w") as f:
-                f.write(ureport)
-
-            self.log_debug("Written to {0}".format(filename))
-            self.known.add(report)
+        try:
+            for report in sorted(reports):
```
sorki: `for i, report in enumerate(sorted(reports)):`

mbrysa:
Thanks, I changed the counter to `enumerate`.

Issue #411, opened 2015-03-24
Primary key of BzBug/MantisBug should be composed of bug's id and tracker_id


sorki:
Outlined in: https://github.com/abrt/faf/pull/409#discussion_r27006402

PR #452, opened 2015-08-21
pyfaf: fix update-bugs for MantisBT

mbrysa:
Related to the fact that MantisBT bugs' IDs are stored in the `external_id`
field not `id` as it is the case with Bugzilla bugs.
Related to #411


sorki:
Thank you!

PR #454, opened 2015-08-25
webfaf2: fix saving of invalid uReports

mbrysa:
`commit` must be used insted of `flush` to save the DB session
inside a web request.


sorki:
Thanks!

PR #455, opened 2015-08-25
webfaf2: add "FIXED" status for problem bugs

### LINKED ISSUE #429: problem "state" column should have multiple CLOSED states
When searching for problems on https://retrace.fedoraproject.org/faf/, there is a "State" column.

It is slightly misleading, because it will display a green "CLOSED" if a bug has been closed in bugzilla.redhat.com in any way, including resolutions like CANTFIX or INSUFFICIENT_DATA that indicate the bug was not actually fixed (we use those often when the bugzilla reporter does not provide steps to reproduce).

The information is available by hovering over the individual bugzilla links in the next column, but that is quite inconvenient, there can be 5 or more bugzilla bugs per problem.

Please add another state there so we can differentiate between "closed-and-fixed" and "closed-but-still-buggy", so that "closed-and-fixed" is only displayed when there is at least one bug with ERRATA or NEXTRELEASE (or similar) resolution.

mbrysa:
Problem will be marked "FIXED" iff all bugs associated with it are closed and
at least one has resolution NEXTRELEASE, CURRENTRELEASE, RAWHIDE or ERRATA.
Closes #429

### DIFF
--- a/src/pyfaf/storage/problem.py
+++ b/src/pyfaf/storage/problem.py
@@ -72,7 +72,13 @@ class Problem(GenericTable):
         if not bugs:
             return 'NEW'
 
-        return sorted(bugs, key=lambda x: x.order())[0].status
+        s = sorted(bugs, key=lambda x: x.order())[0].status
+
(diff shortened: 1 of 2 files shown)


sorki:
Nice, thanks!

PR #457, opened 2015-08-28
pyfaf: load Report bug backrefs based of bugtracker plugins

mbrysa:
Bugtracker plugins are no longer hardcoded and show automatically
in report.item and problem.item and list.


sorki:
Thanks!

PR #462, opened 2015-09-23
Web2 to web rename

sorki:


jfilak:
It looks good!

How can I install FAF from the "web2_to_web_rename branch" using Ansible? I want to test if everything works before merging to master.

Please excuse my ignorance, but  why do we need 3 commits in this pull request?

sorki:
Use 'web2_to_web_rename' branch from ansible repo, set

PR #467, opened 2015-10-15
URL attachment saving

sorki:
Also adds attachment filtering support.


jfilak:
It looks good! Especially the filtering part. Thank you!

mhabrnal:
Looks good to me.

PR #478, opened 2015-11-19
add hash-paths action

sorki:
For replacing private part of the path with hash.


sorki:
Updated.

jfilak:
Good job. Thanks!

PR #479, opened 2015-11-23
old packages garbage collector

sorki:
Removes LOBs for obsolete packages.

### DIFF
--- a/faf.spec.in
+++ b/faf.spec.in
@@ -595,6 +595,7 @@ fi
 %{python_sitelib}/pyfaf/actions/opsysadd.py*
 %{python_sitelib}/pyfaf/actions/opsysdel.py*
 %{python_sitelib}/pyfaf/actions/opsyslist.py*
+%{python_sitelib}/pyfaf/actions/packagegc.py*
 %{python_sitelib}/pyfaf/actions/releaseadd.py*
 %{python_sitelib}/pyfaf/actions/releaselist.py*
 %{python_sitelib}/pyfaf/actions/releasemod.py*
--- a/src/pyfaf/actions/Makefile.am
+++ b/src/pyfaf/actions/Makefile.am
@@ -24,6 +24,7 @@ actions_PYTHON = \
     find_crash_function.py \
     init.py \
     mark_probably_fixed.py \
+    packagegc.py \
     pull_associates.py \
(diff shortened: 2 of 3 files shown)


sorki:
An example would be?

What comes to my mind are some packages that are avoiding rebuilds between releases e.g. packages from fc21 appearing in fc22 or fc23 but i don't know why this actually happens.

jfilak:
Yea, that's the case! And it just happens, @mbrysa was in contact with Fedora relengs. Maybe he has mada a note about it somewhere.

jfilak:
I've found this "old package garbage collector" scripted developed by @mbrysa Unfortunately I'm not able to merge it with your script :(

```
from pyfaf.storage import *
from pyfaf.opsys import systems

db = getDatabase()

# The oldest fedora release you want to keep
new_fedora = 21

old_fedoras = range(1, new_fedora)

fedora = systems["fedora"]

koji_builds = fedora.get_released_builds(str(new_fedora))

koji_builds_set = set(("{}*{}*{}*{}".format(b["name"], b["epoch"], b["version"], b["release"]) for b in koji_builds))

# Deleting LOBs for older Fedoras while keeping builds that appear in the newest one
# i.e. an fc20 build might make it's way into fc21, so we want to keep it
for of in old_fedoras:
    for b in db.session.query(Build).filter(Build.release.like("%fc{0}".format(of))):
        if "{}*{}*{}*{}".format(b.base_package_name, b.epoch, b.version, b.release) in koji_builds_set:
            print("Keeping {0}".format(b.nevr()))
        else:
            print("Deleting build {0}".format(b.nevr()))
            for p in b.packages:
                print("  Deleting package {0}".format(p.nevra()))
                if not p.has_lob("package"):
                    print("  Lob not found")
                    continue
                try:
                    p.del_lob("package")
                except Exception as e:
                    print("Exc: {}".format(e))
```

jfilak:
```
<jfilak> Hello, I would like to know if a Fedora 22 build, let's say foo-0.1-1.fc22, can make it to Fedora 24 repo?
<kalev> jfilak: yes, it can; for example it has failed in F23 and F24 mass rebuilds then the old F22 build is just carried over
```

sorki:
Obsolete, see #528

PR #486, opened 2016-02-08
Fixing bug for setup FAF

patriczek:
Complete saving url attachement to database
Improved repoassign for rhel

### DIFF
--- a/src/pyfaf/actions/repoassign.py
+++ b/src/pyfaf/actions/repoassign.py
@@ -45,10 +45,19 @@ class RepoAssign(Action):
             arch = get_arch_by_name(db, item_name)
 
             if not (opsys or arch):
-                self.log_error("Item '{0}' not found"
+                #If name is rhel we will search Red Hat Enterprise Linux
+                if item_name == "rhel":
+                    item_name = "Red Hat Enterprise Linux"
+                    opsys = get_opsys_by_name(db, item_name)
+                    
+                    if not opsys:   
+                        self.log_error("Item '{0}' not found"
                                .format(item_name))
-
-                return 1
+                        return 1
+                else:
+                    self.log_error("Item '{0}' not found"
+                               .format(item_name))
+                    return 1
 
             if opsys:
                 opsys_list.append(opsys)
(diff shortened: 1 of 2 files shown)


trapas:

sorki:
Should be two commits with proper explanation what each does and why it is required.

src/pyfaf/actions/repoassign.py
```diff
@@ -41,8 +41,15 @@ def run(self, cmdline, db):
         opsys_list = []
 
         for item_name in cmdline.OPSYS + cmdline.ARCH:
+
```
sorki: No need for a blank line

src/pyfaf/ureport.py
```diff
@@ -521,6 +521,11 @@ def save_attachment(db, attachment):
         db_url.report = report
         db_url.url = url
         db_url.saved = datetime.datetime.utcnow()
+        
+        try:
+            db.session.flush()
+        except IntegrityError:
+            raise FafError("Unable to save attachment")
```
sorki: 'Unable to save URL'

src/pyfaf/actions/repoassign.py
```diff
@@ -41,8 +41,15 @@ def run(self, cmdline, db):
...
         for item_name in cmdline.OPSYS + cmdline.ARCH:
+            
             opsys = get_opsys_by_name(db, item_name)
             arch = get_arch_by_name(db, item_name)
+            
+            #If name is rhel we will search Red Hat Enterprise Linux
+            if not (opsys or arch):
+                if item_name == "rhel":
+                    item_name = "Red Hat Enterprise Linux"
+                    opsys = get_opsys_by_name(db, item_name)
```
sorki: This doesn't seem right at all. Fixing rhel -> Red Hat Enterprise Linux ambiguity in one place (case) only?

sorki: This indeed seems like a only place where it's needed.

So please just move the block bellow the following 'if' as the main condition is the same in both clauses.

sorki:
@trapas add to whitelist

PR #488, opened 2016-02-11
Attachments: One url can be attached to many reports

patriczek:
One url can be attached to many reports, but every report must have unique url's
QA'S want save one url for more reports.

Here will be a redundancy, but i think its better then creating junction tables
and join more tables is more expensive on resources.


sorki:
After reading this several times I still don't understand why it is necessary. Simple joins are not expensive btw.

jfilak:
@patriczek Oops, you have removed the commit. Please, address @sorki comment, push the commit to a new branch and open a new pull request.

PR #500, opened 2016-04-22
Attachments: One url can be attached to many reports

patriczek:
Currently, a single URL can be added to the database only once - in other words, it is not possible to attache a single URL to more micro reports. I was not able to find a justification for it.

We have found a use case where we need to attach a single URL to more micro reports (test case URL). So this commit removes the constraints of uniqueness of URL column which is enforced in the application logic.

The constraint is enforced by looking up for the url in the reporturls. The commit adds a new filter condition which will ensure that a single URL can be attached to more micro reports but just only once to a particular one.

### DIFF
--- a/src/pyfaf/ureport.py
+++ b/src/pyfaf/ureport.py
@@ -510,8 +510,11 @@ def save_attachment(db, attachment):
     elif atype == "url":
         url = attachment["data"]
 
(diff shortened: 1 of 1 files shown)


sorki:
Thanks. Wasn't clear before.

PR #507, opened 2016-05-18
Enable usage of yum repositories on systems without yum

jfilak:
We don't need to mess with system yum, we can just download and parse createrepo data.


jfilak:
Yay, I forgot to clean the cache.

PR #509, opened 2016-05-30
Hook RpmMetadata in FAF database

jfilak:
This pull request makes changes necessary for configuring repo instances with the 'rpmmetadata' type.


sorki:
Thanks

PR #515, opened 2016-06-23
pyfaf: use static dictionary for signal name

pavlix:
Rationale: Always explicitly choose the canonical signal name where
multiple names match one signal number. For example always use SIGABRT
as canonical name and not its obscure SIGIOT alias.


trapas:

sorki:
Thanks!

PR #516, opened 2016-06-28
java: drop all requirements on frame name

jfilak:
For the same reason we do not check format of C++ function names, it is
too complex.

Signed-off-by: Jakub Filak <jfilak@redhat.com>


sorki:
I guess that test failure is not related. merging.

PR #519, opened 2016-07-07
ureport: stop wasting memory by loading all ureports

jfilak:
The removed command is clearly a bug. There is no need to load all
micro-reports and the variable is never used.

Signed-off-by: Jakub Filak <jfilak@redhat.com>


sorki:
Good catch. Thanks

PR #528, opened 2016-09-20
Add old package garbage collector

marusak:
We need to be able get all package of a single release to perform efficient garbage collection of packages/builds of EOL releases. Now it is not possible to get all builds that belong to a specific operating system release because there is no relationship between operating system releases and builds. To make this possible we must introduce a new link table that will connect Builds, OpSysRelease and Arch. Having this link table would also solve another problem and that is need to be able to create a yum repository from all builds of a single release of operating system in order to be able to install a minimal set of packages to prepare a chroot environment for backtrace generation. 

The table will be populated in the “reposync” action. The action iterates over all packages in a repository denoted by URL and creates new builds. At this time we must create the connection between the new build and operating system release.

There are currently two types of repositories:
Repos assigned to a specific operating system (so called parametrized repositories)
Pure repo without relations

We need to introduce a third type of repository - repository assigned to operating system releases - which will be similar to the parametrized repositories but its URL will not include any parameters ($..).

Assigning operating system release to a third type of repository could be done by modifying  existing “repoassign” action. This module is now responsible for assigning operating systems to a repositories of the first type. Each assigned operating system is saved into OpSysRepo. To be able to save also releases, we need to introduce another table - OpSysReleaseRepo. Then “repoassign” action would have to decide, which table it will fill. 
This can be done by simple parser, that will try to read release from passed arguments. If the parser succeeds, a new item will be saved into  OpSysReleaseRepo rather than into OpSysRepo.
Example:
Resolves as 1st type: `faf repoassign repo_name Fedora x86_64 i686`
Resolves as 3rd type: `faf repoassign repo_name “Fedora 24” x86_64 i868’
(Note: since there is infinite amount of ways, in which opsys with release can be written, it would be sensible to make it strict - operating system is separated by one space from release. Also the operating system should start with capital letter.)

Currently “reposync” before adding any packages or builds goes through all existing repositories and checks theirs types. This check is done by testing if dollar sign is present in a url link. If is, the repository is considered the first type, otherwise the second. After adding the third type, this check would not be correct, since it would not be able to distinguish between second and the third type. Therefore a new condition is needed. It should be possible to check, which table from OpSysReleaseRepo or OpSysRepo is available. This way it could be possible to tell apart 1st, 2nd and 3rd version. (First version has OpSysRepo second has none and third has OpSysReleaseRepo)

In this stage “reposync” action would be able to read from repository operating system and release, if they are present. (Note: First type also has release in this stage. Dollars are replaced by active releases). And therefore when a new build is being added, we can assign it a operating system and release. Since one build can be for more than one operating system or release and even architectures a link table must be introduced. This table will connect Builds, OpSysRelease and Arch and therefore a suitable name would be BuildOpSysReleaseArch.

It is worth mentioning, that now it is necessary, that to each repository at least one architecture is assigned, since from now on architectures are read only from repositories and not from packages, as it was until now.

Finally a efficient garbage collector could be added.

### DIFF
--- a/faf.spec.in
+++ b/faf.spec.in
@@ -229,6 +229,13 @@ Requires: %{name} = %{faf_version}
 %description action-find-components
 A plugin for %{name} implementing find-components action
 
+%package action-assign-release-to-builds
+Summary: %{name}'s assign-release-to-builds plugin
+Requires: %{name} = %{faf_version}
+
+%description action-assign-release-to-builds
+A plugin for %{name} implementing assign-release-to-builds action
+
 %package action-find-crash-function
 Summary: %{name}'s find-crash-function plugin
 Requires: %{name} = %{faf_version}
@@ -830,6 +837,9 @@ fi
 
 %files action-find-report-solution
 %{python_sitelib}/pyfaf/actions/find_report_solution.py*
+
+%files action-assign-release-to-builds
+%{python_sitelib}/pyfaf/actions/assign_release_to_builds.py*
 
 %files action-repo
 %{python_sitelib}/pyfaf/actions/repoadd.py*
--- a/faf.spec.in
+++ b/faf.spec.in
@@ -358,6 +358,13 @@ Requires: %{name}-fedmsg = %{faf_version}
 %description action-fedmsg-notify
 A plugin for %{name} implementing notification into Fedmsg
 
+%package action-cleanup-packages
+Summary: %{name}'s cleanup-packages plugin
+Requires: %{name} = %{faf_version}
+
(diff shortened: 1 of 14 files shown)


trapas:

jfilak:
add to whitelist

jfilak:
ok to test

patriczek:
it works for me

sorki:
Well done. Rebase and merge it!

jfilak:
Please, update also the alembic version script.

jfilak:
Thank you!

PR #535, opened 2016-10-24
HTTP API for statistic aggregator

patriczek:
In this pull request i made a HTTP API which provides data for our statistic aggregator.


sorki:
LGTM. If you plan to add more json API stuff it might be better to introduce automatic serialization to DeclarativeBase which all models inherit instead of writing serialize manually (something like https://github.com/kolypto/py-flask-jsontools#jsonserializablebase or better https://marshmallow-sqlalchemy.readthedocs.io/en/latest/).

PR #539, opened 2016-11-22
pyfaf: Fix revision tree

marusak:
In the commit e88e67d a new migration script was added but its downversion did
not point to the current head. That meant that there were two heads.

Before this commit:
    $alembic heads
        >133991a89da4 (head)
        >183a15e52a4f (head)
After this commit:
    $alembic heads
        >183a15e52a4f (head)


jfilak:
Thank you!

PR #546, opened 2016-11-28
Support mirrors

marusak:
This pull-request enables repositories to have more than one url, so in case that
one url is not available in time of reposyncing, another can be used.

### DIFF
--- a/src/pyfaf/actions/repoimport.py
+++ b/src/pyfaf/actions/repoimport.py
@@ -21,7 +21,7 @@ import ConfigParser
 
 import pyfaf.repos
 from pyfaf.actions import Action
-from pyfaf.storage.opsys import Repo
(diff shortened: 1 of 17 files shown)


src/pyfaf/actions/repoimport.py
```diff
@@ -99,8 +101,8 @@ def run(self, cmdline, db):
                                .format(repo.name))
                 return 1
 
-            self.log_info("Adding repository '{0}' ({1})"
-                          .format(repo.name, repo.url))
+            self.log_info("Adding repository '{0}' {1}"
+                          .format(repo.name, [url.url for url in repo.url_list]))
```
jfilak: Nit: What about to convert the list of URLs to string? The default conversion looks terrible.

Issue #543, opened 2016-11-22
migration problem after update


mjtrangoni:
I made an RPM update yesterday from master branch to our private retrace server, and I am getting some troubles with this. I also discover  'faf-migrate-db' afterwards :-1: 

The question is, how can I get this working again? I click on some Problems reports and the Web-UI is working, but others are not and generate this python crash: 

```
Exception on /reports/124/ [GET]
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 1381, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/lib/python2.7/site-packages/webfaf/reports.py", line 495, in item
    unique_ocurrence_os[os_name]['unique'] += ch.unique
TypeError: unsupported operand type(s) for +=: 'NoneType' and 'NoneType'
```
From `create-problems` I am getting 
```
$ faf create-problems >> /var/log/faf/create-problems.log                                                                                                                    
INFO:faf.CreateProblems:[1 / 5] Processing problem type: Unhandled Python exception
INFO:faf.CreateProblems:Creating problems from clusters
INFO:faf.CreateProblems:[2 / 5] Processing problem type: Kernel oops
INFO:faf.CreateProblems:Creating problems from clusters
INFO:faf.CreateProblems:[3 / 5] Processing problem type: Unhandled Ruby exception
INFO:faf.CreateProblems:No reports found
INFO:faf.CreateProblems:Creating problems from clusters
INFO:faf.CreateProblems:[4 / 5] Processing problem type: Unhandled Java exception
INFO:faf.CreateProblems:No reports found
INFO:faf.CreateProblems:Creating problems from clusters
INFO:faf.CreateProblems:[5 / 5] Processing problem type: Crash of user-space binary
WARNING:faf.CoredumpProblem:Report #549 has only one bad frame
[...]

PR #549, opened 2016-12-14
Set default value in alembic revision and histories model

patriczek:


sorki:
Seems to revise very old version, please rebase. Also make sure to add it to src/pyfaf/storage/migrations/versions/Makefile.am

marusak:
This is not sufficient. It does not set "0" to already empty columns. It is important to fill columns with zeros. Also add this to the table definition in src/pyfaf/storage/report.py.

marusak:
Now it works. Also fixes #543

PR #552, opened 2017-02-13
Development/debugging tools

mcyprian:
In this PR:
   - simple manage.py script providing local development server and shell features, more commands 
     for testing, db management etc. can be added later

    - requirements.txt file specifying set of dependencies, unfortunately some requirements
      aren't available on PyPI. A few RPMs must be installed on the system and visible inside
      virtualenv created for FAF.

It is possible to run development server and shell directly from sources thanks to this tools.
To make it completely independent of installed instance of FAF it is necessary to create config file
similar to [this example](https://paste.fedoraproject.org/557383/) in the local directory and set
FAF_CONFIG_FILE environment variable to point to this config file.

### DIFF
--- a/src/webfaf/manage.py
+++ b/src/webfaf/manage.py
@@ -1,0 +1,24 @@
+#!/usr/bin/env python2
+
+import os
+import sys
+
+from flask_script import Manager, Shell
+
+sys.path.insert(0, os.path.dirname(os.path.abspath(os.path.dirname(__file__))))
+
+from webfaf_main import app, db
+from webfaf_main import import_blueprint_plugins
+
+
+import_blueprint_plugins(app)
+manager = Manager(app)
+
+
+def make_shell_context():
+    return dict(app=app, db=db)
+
+manager.add_command('shell', Shell(make_context=make_shell_context))
+
+if __name__ == '__main__':
+    manager.run()
--- a/config/faf-devel.conf
+++ b/config/faf-devel.conf
@@ -1,0 +1,18 @@
+# Faf development configuration file
+[Main]
+PluginsDir = plugins/
+TemplatesDir = templates/
+AutoEnablePlugins = True
+
+[Storage]
+ConnectString = postgresql:///faf
+LobDir = ~/.local/var/spool/faf/lob
+
+[uReport]
+# The directory that holds 'reports' and 'attachments' subdirectories
+Directory = ~/.local/var/spool/faf
+CreateComponents = False
+# attachments accepted by this server
+# allowed values: fedora-bugzilla rhel-bugzilla centos-mantisb comment email url
+# or * to allow all attachments
+AcceptAttachments = fedora-bugzilla rhel-bugzilla centos-mantisbt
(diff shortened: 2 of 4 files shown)


trapas:

jfilak:
add to whitelist

jfilak:
ok to test

jfilak:
test this please

marusak:
Thanks, it is a great idea to have this. But there are some lacks, that I think would be great to have.

1. When the variable FAF_CONFIG_FILE is not set, it would be lovely if the script generated a default one. For example, created directory `/var/tmp/development-faf-dir` and into file /`var/tmp/development-faf-config` wrote a default config pointing into `/var/tmp/development-faf-dir`.

2. When there is no Postgres faf database on local host, this will not work. It excepts the developer to have a valid and functional database running. Also it expects that the executing user of `manage.py` has all rights to access this DB. I believe it would be very useful to have some dummy DB ready, that would contain some data. This DB could probably be a SQLite one. I dont think it is suitable to have a DB dump on github, but put it somewhere publicly available.

3. Support also faf's actions. It should be possible to add another command into `manage.py` that would execute also faf's actions.

mcyprian:
This is of course only the first step on a long journey to complex and easy to use environment for developers. All three listed features will be useful, I agree. I can take a look at first two of them now.

marusak:
The third could be something like this. It is a working code, just not sure about the popen...
https://paste.fedoraproject.org/557955/

marusak:
Ok, so my previous code worked only because I had installed faf. When I removed it, it stopped working.
https://paste.fedoraproject.org/557990/

src/webfaf/manage.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/python
```
marusak: Change this please to `#!/usr/bin/env python`  so it can be run inside virtualenv like ./manage.py

jfilak: But FAF is a Python2 application, right?

marusak: It is, so it should be `#! /usr/bin/env python2`

marusak:
@trapas retest

marusak:
retest this please

sorki:
Adding requirements with fixed versions seems like a bad idea, you can run into inconsistencies as by default it uses packages from Fedora. Also who will maintain that file?

Also there's an easy way to set-up faf & retrace https://github.com/abrt/abrt-ansible#using-vagrant

sorki:
If you want to have this installed on your machine instead of a VM 'dnf install faf' will install all the dependencies. Then you just need to point your PYTHONPATH to cloned repository.

marusak:
@sorki This is for development and debugging. So you can run manage.py in terminal and on other tab you can edit files in cloned repository and only refresh locally deployed  webpages. Also it prints executed sql commands, you can easily add actions so debugging and testing faf is crazy easy. The ansible you mentioned is great for deploying into production. The idea with pointing PYTHONPATH into cloned repository sound interesting, but I am not sure if it provides webserver, prints debugging information and executing all action will work. If so, I will definitely try it!

mcyprian:
@sorki If you don't like the idea of requirements.txt file, which needs to be updated form time to time as you pointed out, manage.py can fully rely on system installed dependencies from RPMs. Benefit of having manage.py is in my opinion still significant.

sorki:
Yeah sure. No problem with manage.py, even better if you try to integrate this into faf cli, see for example faf shell action https://github.com/abrt/faf/blob/master/src/pyfaf/actions/shell.py that launches a ipython shell with faf imported + some shortcuts.

PR #554, opened 2017-02-14
webfaf: Always show executable on report page

### LINKED ISSUE #524: Always show executable on the Report page
If stack trace does not include main, it is hard to guess what executable actually crashed.

marusak:
If stack trace does not include main, it is hard to guess what
executable actually crashed. After this commit, executable is always
displayed.

Fixes #524

### DIFF
--- a/src/webfaf/reports.py
+++ b/src/webfaf/reports.py
@@ -34,6 +34,7 @@ from pyfaf.storage import (AssociatePeople,
                            ReportHistoryMonthly,
                            ReportUnknownPackage,
                            ReportBacktrace,
+                           ReportExecutable,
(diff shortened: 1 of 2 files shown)


jfilak:
We should develop automated test for Web UI! I really miss a test case definition in this PR.

marusak:
We definitely should. But I am not sure, how much effort it will take.
p.s. the build failed because the lack of space for jenkins, not a really bug.

marusak:
retest this please

sorki:
Web part is already tested, you just need to write an actual test(s), like in https://github.com/abrt/faf/blob/master/tests/webfaf/reports

PR #555, opened 2017-02-17
speed up assign_release_to_builds

xsuchy:
if we do not commit after each package it is ~1000x faster

### DIFF
--- a/src/pyfaf/actions/assign_release_to_builds.py
+++ b/src/pyfaf/actions/assign_release_to_builds.py
@@ -28,6 +28,7 @@ class AssignReleaseToBuilds(Action):
 
     def __init__(self):
         super(AssignReleaseToBuilds, self).__init__()
+        self.uncommited = 0
 
     def run(self, cmdline, db):
         # nobody will write the full name
@@ -96,6 +97,7 @@ class AssignReleaseToBuilds(Action):
                 self._add_into_build_opsysrelease_arch(db, build,
                                                        opsysrelease, arch)
 
+        db.session.flush()
 
     def _edit_opsys(self, original_name):
         """ Solve name problem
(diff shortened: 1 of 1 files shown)


trapas:

marusak:
ok to test

marusak:
add to whitelist

marusak:
@trapas test this please

marusak:
test this please

src/pyfaf/actions/assign_release_to_builds.py
```diff
@@ -83,6 +83,7 @@ def run(self, cmdline, db):
                 if found_build:
                     self._add_into_build_opsysrelease_arch(db, found_build,
                                                            opsysrelease, arch)
+                db.session.flush()
```
marusak: Shouldn't this be one level up? It is in the for loop, wasn't the point of this commit to have flush outside loops?

jfilak:
What about to call flush after 1000 iterations? I'm not sure it is safe to commit huge data. Do you know why we commit data after each loop?

src/pyfaf/actions/assign_release_to_builds.py
xsuchy: It should. I will change it.

xsuchy:
@jfilak Saving data after each loop have sense if those data are expensive to extract, so you have safepoint, from which you can restart.
However these data are easy to extract. The original code run over night and quick interpolation was that it may run for another 45 days. More or less. After this change, the operation was finished after lunch break.
And yes flush after 1000 iteration can be done too. But that commit is not huge. Each for-loop consist only from several thousands of rows. And the journal is not big (just few MB) and easily fit in memory.

marusak:
I've updated this PR so it flushes after 1000 iterations or before the action finishes.

marusak:
ok to test

PR #560, opened 2017-03-02
config: Change WSGI process group

mkutlak:
This change will allow to run multiple applications that use mod_wsgi
on one host.


sorki:
LGTM, thanks.

PR #563, opened 2017-03-28
changelog: Update after a while

marusak:
I tried to select some bigger and noticeable changes. But since from
last release there were more 300 commits and I don't really know what
was happening back then, I just selected what was clear to me.

I hope from now on the changelog will updated regularly.


marusak:
retest this please

PR #566, opened 2017-04-19
Fix url to FAF report in FedMsg notification

### LINKED ISSUE #564: Fedmsg notifications contain not working URL to FAF
Fedmsg notifications sent by FAF look like the following:
https://apps.fedoraproject.org/datagrepper/id?id=2017-71bd9a69-c8c9-4bfe-90ed-429245492eb1&is_raw=true&size=extra-large

``url`` in ``msg`` part of the notification starts with ``http://...`` but the ``retrace.fedoraproject.org`` server is running over HTTPS.

The issue is here:
https://github.com/abrt/faf/blob/master/src/pyfaf/utils/web.py#L54
url_for() returns URL prefixed with ``http://`` even if the server is running over HTTPS.

We should define ``app.config['PREFERRED_URL_SCHEME'] = 'https'`` if the server is running over HTTPS.

mhabrnal:
URLs attached into FedMsg notifications were always prefixed
by 'http://' even the server runs over https.
The URL prefix is now controlled by a new option 'require_https' which
was added into 'web.conf'. If the option is 'true' URLs are prefixed
by 'https://'.

Fixes #564

Signed-off-by: Matej Habrnal <mhabrnal@redhat.com>

### DIFF
--- a/config/plugins/web.conf
+++ b/config/plugins/web.conf
@@ -2,6 +2,7 @@
 debug = false
 proxy_setup = false
 secret_key = @SECRET_KEY@
+require_https = true
 url = https://example.org/faf/
 # server name is the bare URL without protocols and trailing slash
 server_name = example.org/faf
--- a/src/pyfaf/utils/web.py
+++ b/src/pyfaf/utils/web.py
@@ -39,6 +39,19 @@ def server_name():
         return None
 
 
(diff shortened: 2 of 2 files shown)


trapas:

mhabrnal:
test this please

PR #574, opened 2017-05-16
Codestyle fix

TechMarinerh:
Fixing codestyle as internship in redhat.

### DIFF
--- a/src/pyfaf/ureport.py
+++ b/src/pyfaf/ureport.py
@@ -594,10 +594,9 @@ def is_known(ureport, db, return_report=False, opsysrelease_id=None):
         found = True
 
     elif ('BUG_OS_MINOR_VERSION' in known_type and
-        get_reportbz(db, report.id, opsysrelease_id).first() is not None):
+          get_reportbz(db, report.id, opsysrelease_id).first() is not None):
 
         found = True
-
     elif ('BUG_OS_MAJOR_VERSION' in known_type and
           get_reportbz_by_major_version(db, report.id,
                                         major_version=ureport["os"]["version"]
--- a/src/pyfaf/ureport.py
+++ b/src/pyfaf/ureport.py
@@ -582,7 +582,8 @@ def is_known(ureport, db, return_report=False, opsysrelease_id=None):
     if 'EQUAL_UREPORT_EXISTS' in known_type:
         report_os = ureport["os"]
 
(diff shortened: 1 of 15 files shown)


trapas:

marusak:
Thank you for your work!
Let me test this.
Also I don't think that 'bordel' is a suitable commit message:) However I don't think it a good idea to create 95 oneline commits. I will squash them into one. @xsuchy do you agree? - it blurs history terribly.

marusak:
test this please

xsuchy:
test this please

xsuchy:
PR fixed. 
@marusak quite opposite. This will give for every change the reason why it was done. And if it cause problem we can revert only that one problem. E.g. Right now I told @TechMarinerh to remove the commits where he removed parenthesis after print, which will cause problems when we will migrate to python3

jfilak:
Oh ... Please squash it as much as possible. If you expect problems, then don't call this PR "style fixes".

xsuchy:
test this please

marusak:
test this please

xsuchy:
@marusak @jfilak similar issues in the same file squashed together. I hope you like it now.

marusak:
9ce0dff and 40318fe both edit Wrong continued indentation in clenaup_packages and are not squashed together for example. And the names are confusing as well ( (add 5 spaces) even though 5 and 6 spaces are added and (add 1 space) even though some spaces are added or removed.

I would squash a bit more and write messages like "Wrong indentation" instead of "Wrong (hanging|continued) indentation. 

This was only one example, it is possible to squash a lot more though:) I will be happy with 5 or so commits:)

And I still think that not potentially problematic stuff (there are almost none) such as removing spaces in function calling or in long expression broken over a few lines could be easily one commit for all files.

src/pyfaf/ureport.py
```diff
@@ -582,7 +582,8 @@ def is_known(ureport, db, return_report=False, opsysrelease_id=None):
     if 'EQUAL_UREPORT_EXISTS' in known_type:
         report_os = ureport["os"]
 
-    report = get_report(db, report_hash, os_name=report_os['name'], os_version=report_os['version'], os_arch=report_os['architecture'])
+    report = get_report(db, report_hash, os_name=report_os['name'],
+ os_version=report_os['version'], os_arch=report_os['architecture'])
```
marusak: This products another warning `[bad-continuation] Wrong continued indentation (add 24 spaces).`

PR #575, opened 2017-05-17
Codestyle fix 2

TechMarinerh:
Fixing code style as internship in redhat.

### DIFF
--- a/src/pyfaf/actions/assign_release_to_builds.py
+++ b/src/pyfaf/actions/assign_release_to_builds.py
@@ -138,7 +138,7 @@ class AssignReleaseToBuilds(Action):
 
             db.session.add(bosra)
             self.uncommited += 1
-            if (self.uncommited > 1000):
+            if self.uncommited > 1000:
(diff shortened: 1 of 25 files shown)


trapas:

src/pyfaf/actions/assign_release_to_builds.py
```diff
@@ -57,29 +57,29 @@ def run(self, cmdline, db):
...
             for build in systems[opsys].get_released_builds(cmdline.RELEASE):
                 found_build = (db.session.query(Build)
-                         .filter(Build.base_package_name ==
-                                 build["name"])
-                         .filter(Build.version == build["version"])
-                         .filter(Build.release == build["release"])
-                         .filter(Build.epoch == build["epoch"])
-                         .first())
+                               .filter(Build.base_package_name ==
+                                       build["name"])
```
marusak: this break is not really necessary. It is only 81 characters without break - when you rebase, the warning goes away:)

PR #577, opened 2017-06-21
webfaf: remove redefinitions of built-in list function

mcyprian:
Modules reports, problems and dumpdirs contain functions called `list()`. I don't consider redefinition of Python's built-in functions to be a good practice. This can cause an errors that are really hard to debug. I suggest to rename these functions. In my opinion the name "dashboard" reflects the data content of documents on routes /problems and /reports.

### DIFF
--- a/src/pyfaf/actions/stats.py
+++ b/src/pyfaf/actions/stats.py
@@ -298,7 +298,7 @@ class Stats(Action):
             out += "\n"
             if webfaf_installed():
                 out += "URL: "
-                out += reverse("problems.list")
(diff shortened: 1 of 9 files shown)


marusak:
Thank you for PR.
Looks good, however I believe that this should be two separate commits (one for the list stuff and another for the unnecessary imports). Could you split it?

Also https://github.com/abrt/faf/blob/master/src/webfaf/dumpdirs.py#L235 shouldn't this be also changed?

mcyprian:
Make sense, it is separated now.

PR #580, opened 2017-07-04
Create problems speedup

marusak:


src/pyfaf/actions/create_problems.py
```diff
@@ -341,66 +350,145 @@ def _create_problems(self, db, problemplugin, report_min_count=0):
...
+                    for db_report in problem:
+                        if db_report.problem_id:
+                            origin_report = db_report
+
+                    # Problem created only from new reports
+                    comps = {}
+                    if not origin_report:
+                        new = Problem()
+                        db.session.add(new)
+                        db.session.flush()
```
xsuchy: why is flush() here. I am asking because on line 423 you are flushing it again, which makes more sense to me.

marusak: because I need a ID on line 374. Without flushing I don't have the ID.

PR #581, opened 2017-07-07
Problems dashboard improvement

mcyprian:
The main problems query is limited to the first 40 results ordered by count in case filters are not specific enough. There is a "next page" pagination button moving to next 40 results, but it doesn't provide very flexible way to browse problems, especially when the table contains thousands of rows.

In this PR:
- problems query is rewritten in raw SQL, which brings significant improvement of performance
- number of rows is not limited
- streaming from template is used instead of rendering
- DataTables library provides pagination, instant search and multi-column ordering

The database in my test environment contains approximately 17 thousand of problems, I guess this should be tested on real production data to make sure it is fast enough.
All the filters were implemented according to the old implementation of the query using ORM. 
I haven't manage to implement "Tainted" label, its calculation seems to be very complicated (properties of 3 objects of ORM model calling each other). I am not sure how important is this part, filtering based on various taint flags works fine.


xsuchy:
Please fix the failing check.

marusak:
2 comments:

1.  You removed file list_table_rows.html from problems (there is also one in reports, wouldn't it be better to use the same stuff under problems and reports?). You need to edit also specfile and makefile. After that you should be able to run make rpm, which runs tests (and some are failing, please investigate what is the problem).

2. I am not really sure about the rewriting into raw SQL. It makes it very hard to debug, fix and improve. But I would like to hear option of others on this.

jfilak:
The Tainted flag is very important for Kernel folks. Is the rewrite to SQL really necessary or is there just for speed improvements?

marusak:
> Is the rewrite to SQL really necessary or is there just for speed improvements?

It is for the speed improvement. You know, when you open for example https://retrace.fedoraproject.org/faf/problems it should load much faster.  _(But when I thing of it now, for reports it would be more useful (since there is more of them and also /sumarry which is defualt page uses reports))_

What is your take on this? Do you agree with rewriting to raw sql for speed improvement?

jfilak:
> It is for the speed improvement.

OK, so why it is so slow? The answer that "it is written in ORM" is not good enough, IMHO. We should find the real cause - like it requires to load all detail records which are not needed. Because I believe that SQLAlchemy translates the query into pretty much the same SQL command as Michal has created.

> What is your take on this? Do you agree with rewriting to raw sql for speed improvement?

If there is no other way, I would rather create SQL views instead of huge SQL commands. It's easier to debug a view than several lines long command you need to copy'n'paste.

xsuchy:
> OK, so why it is so slow? The answer that "it is written in ORM" is not good enough,

It is because of lazy loading. When you have table with 1000 rows, the ORM will create query which will produce 1000 records, but relations to other tables are not fetched, they are loading only when you access that attribute. So when you want to print whole table with 1000 rows (using patternfly) then you end up with 1+1000*N (where N is number of relations from main table) queries. Those queries are very simple, but there is simply too many of them.
You can disable lazy loading - theoretically, I was not able to do that in real world when we done similar stuff for Copr and Satellite.

xsuchy:
We are fine with this when you rebase and fix the building and failing tests.

marusak:
@mcyprian Can we expect update of this PR?

mcyprian:
I've made some progress but there is a problem with [to_semver](https://github.com/abrt/faf/blob/master/src/pyfaf/storage/custom_types.py#L79) method which in some cases returns different format of version from **to_semver** function defined in PostgreSQL semver extension. I was also testing this implementation on slightly bigger data (20GB dump size) and it seems this is still not fast enough to get all the problems at once for such data. Unfortunately i don't have much time to work on this now.

PR #585, opened 2017-09-08
docker: Introduce FAF docker image

marusak:
Obsoletes abrt/faf-docker repository.

Why have I created a new docker image?

Firstly, because I didn't know there is one.
Secondly, the image needed some rewriting and rethinking.

So why is it directly in FAF repo?
Because it is the correct place. We do not need another repository for
so little data. And also, nobody misses it ever again:)

How does it differ?
The biggest difference is that it builds FAF from currently checked out
git branch. It makes it perfect for development. It makes however, the image a
bit bigger (there are some build-requires).

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,36 @@
+FROM fedora:26
+MAINTAINER mmarusak@redhat.com
+
+ARG password
+
+RUN dnf -y update && \
+    dnf install -y dnf-plugins-core && \
+    dnf -y copr enable @abrt/faf-el7 && \
+    dnf -y install git \
+                   sudo \
+                   which \
+                   cronie \
+                   postgresql-server \
+                   postgresql \
+                   python-psycopg2 \
+                   pg-semver \
+                   faf-* && \
+    dnf clean all && \
+    echo faf:${password:-passwd} | chpasswd && \
+    echo 'faf ALL=(ALL:ALL) NOPASSWD:ALL' >> /etc/sudoers && \
+    echo 'root ALL=(ALL:ALL) NOPASSWD:ALL' >> /etc/sudoers
+
+# Copy initialization script
+COPY docker/files/init_db /usr/bin
+
+# Create crontabs and configurate FAF
+ADD docker/files/crontab-faf /etc/faf/crontab-faf
+RUN crontab -u faf /etc/faf/crontab-faf && \
+    sed -i -e"s/CreateComponents\s*=\s*False/CreateComponents = True/i" /etc/faf/faf.conf && \
+    sed -i -e"s/type\s*=\s*simple/type = null/i" /etc/faf/plugins/web.conf
+
+VOLUME /var/lib/postgres
+VOLUME /var/spool/faf
+
+EXPOSE 80
+CMD ["sh", "-c", "crond && apachectl -D FOREGROUND"]
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,59 @@
+# FAF in Docker container
+
+**Easiest way how to run FAF**
+---
+
+## How to deploy
+
+It is as simple as:
+
+`docker run --name faf -dit abrt/faf-image`
+
+However you also probably want to mount volumes to `/var/lib/postgres` and
+to `/var/spool/faf` not to lose database and FAF's data.
+
+`docker run --name faf -v /var/lib/faf-docker/faf:/var/spool/faf -v
+/var/lib/faf-docker/postgres:/var/lib/postgres/ -dit abrt/faf-image`
+
+If you run FAF for the first time, then there is no database. You have to
+initialize it.
+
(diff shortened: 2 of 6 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM fedora:26
+MAINTAINER mmarusak@redhat.com
+
+# Add copr repo
+ADD docker/files/faf.repo /etc/yum.repos.d/
+
+RUN useradd faf && \
```
xsuchy: Should be: `useradd --system -g faf -d /etc/faf -s /sbin/nologin faf`

docker/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    echo 'faf:passwd' | chpasswd && \
```
xsuchy: Should be:
```
ARG password
RUN sh -c "echo faf:${password:-somepasswordthatyoushouldratherchange} |chpasswd"

File: docker/Dockerfile [hunk 3]
```diff
@@ -0,0 +1,37 @@
+FROM fedora:26
+MAINTAINER mmarusak@redhat.com
+
+# Add copr repo
+ADD docker/files/faf.repo /etc/yum.repos.d/
```
xsuchy: I would rather see: dnf install dnf-plugins-core && dnf copr enable ...

docker/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN crontab -u faf /etc/faf/crontab-faf && \
+    sed -i -e"s/CreateComponents\s*=\s*False/CreateComponents = True/i" /etc/faf/faf.conf && \
+    sed -i -e"s/type\s*=\s*simple/type = null/i" /etc/faf/plugins/web.conf
+
+VOLUME /var/lib/postgres
+VOLUME /var/spool/faf
+
+EXPOSE 80
+CMD ["sh", "-c", "crond; sudo -u postgres /usr/bin/pg_ctl start -D /var/lib/postgres/data -w; \
+                  apachectl -D FOREGROUND"]
```
xsuchy: I would propose:
s/;/&&/g

docker/README.md
```diff
@@ -0,0 +1,59 @@
...
+---
+
+## How to deploy
+
+It is as simple as:
+
+`docker run --name faf -dit abrt/faf-image`
+
+However you also probably want to mount volumes to `/var/lib/postgres` and
+to `/var/spool/faf` not to loose database and FAF's data.
```
xsuchy: s/loose/lose/

docker/README.md
```diff
@@ -0,0 +1,59 @@
...
+Finding out container IP address:
+
+`docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' faf`
+
+Also to send reports into your own FAF, you have to set up libreport on all
+machines, from with you wish to report into your own FAF. To do so, set up
+`URL = http://<container_IP>/faf` in `/etc/libreport/plugins/ureport.conf`.
+
+## Configuring FAF
+New containers comes with fully working and configured FAF (on top of basic configuration
```
xsuchy: s/comes/come/

docker/README.md
```diff
@@ -0,0 +1,59 @@
...
+Fedora releases are added, caching is disabled and FAF accepts unknown components).
```
xsuchy: comma after 'disabled' - It appears that you are missing a comma before the coordinating conjunction and in a compound sentence.

docker/files/init_db
```diff
@@ -0,0 +1,22 @@
...
+mkdir -p /var/lib/postgres/data
+chown postgres:postgres /var/lib/postgres/data
+
+mkdir -p /var/spool/faf/reports/incoming /var/spool/faf/reports/saved /var/spool/faf/reports/deferred /var/spool/faf/lob /var/spool/faf/attachments/incoming
+chown -R faf:faf /var/spool/faf/reports
+chown -R faf:faf /var/spool/faf/lob
+chown -R faf:faf /var/spool/faf/attachments
+
+sudo -u postgres /usr/bin/pg_ctl initdb -D /var/lib/postgres/data > /dev/null
+sudo -u postgres /usr/bin/pg_ctl start -D /var/lib/postgres/data -w > /dev/null
```
xsuchy: should this be called using `systemctl`?

docker/Dockerfile
marusak: useradd: group 'faf' does not exist
Are you sure about the `-g faf` part? 
From useradd man pages: "By default, a group will also be created for the new user"
You probably meant -U `Create a group with the same name as the user, and add the user to this group.`, haven't you?

docker/files/init_db
xsuchy: Note for me, it fails due d-bus. Waiving out.

xsuchy:
Nice. Thank you.

PR #587, opened 2017-09-15
Ureport without packages

marusak:
As respsonse to:
https://github.com/abrt/libreport/pull/503
https://github.com/abrt/abrt/pull/1266
https://github.com/abrt/satyr/pull/262

### DIFF
--- a/config/faf.conf.in
+++ b/config/faf.conf.in
@@ -32,6 +32,10 @@ CreateComponents = False
 # or * to allow all attachments
 AcceptAttachments = fedora-bugzilla rhel-bugzilla centos-mantisbt
 
+# If True accept uReports without packages and with executable that did
+# not come from package
+ProcessUnpackaged = False
+
 # Determines which strategy will be used for searching known or uknown ureport's
 # and bugzilla bug's, if known is empty, then is used BUG_OS_MINOR_VERSION
 #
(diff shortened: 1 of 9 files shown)


jfilak:
I still don't understand why you need a new bool member to store the information you can get by checking other members.

marusak:
Which bool member? In uReport or in database?
In uReport - to know it comes from unpackaged - so then if there are no packages not to be dismiss the uReport and to be able to save in db that it is unpackaged
In DB - To be able to inform user it comes from unpackaged - so he/she does not look for the package

Where can I get this info from - from which members?

jfilak:
> In uReport - to know it comes from unpackaged - ...

No package information denotes unpackaged problem?

> Where can I get this info from - from which members?

component-id? or no package information available?

marusak:
> No package information denotes unpackaged problem?
Well, there are still other packages (those parsed from DSO list). But I now realised that packages are marked as 'affected' and 'related'. I'll look at that if it would be possible to consider report without `affected` package as unpackaged (even though I am a bit sceptical about that).

> component-id? or no package information available?
How could I get it from component id?

jfilak:
> How could I get it from component id?

Null/none/0/''/empty or a well known constant?

marusak:
>Null/none/0/''/empty

Yeah... Component is quite essential. For example it is displayed in tables in reports and problems as well. And also report without any component is useless for users (it is not imediatelly visible where does the problem happened and if I should care). Also I belive that omitting component would cause so many problems, it is not worth it.

>  or a well known constant?

That is better idea, that would not break half of the FAF, but still be quite useless for users.

marusak:
Replaced by #590

PR #590, opened 2017-09-22
Unpackaged from packages

marusak:
As respsonse to:
abrt/libreport#503
abrt/abrt#1266
abrt/satyr#262

### DIFF
--- a/src/webfaf/templates/reports/item.html
+++ b/src/webfaf/templates/reports/item.html
@@ -33,6 +33,9 @@
         {% else %}
           <h3>Info</h3>
         {% endif %}
+        {% if unpackaged %}
+          Note: Executable of this report does not come from a package.
+        {% endif %}
         <dl class='info dl-horizontal'>
           <dt>Problem</dt>
           <dd>
--- a/config/faf.conf.in
+++ b/config/faf.conf.in
@@ -32,6 +32,10 @@ CreateComponents = False
 # or * to allow all attachments
 AcceptAttachments = fedora-bugzilla rhel-bugzilla centos-mantisbt
 
+# Allow uReports without affected package - meaning that crashing code was
+# not packaged
+allow-unpackaged = False
+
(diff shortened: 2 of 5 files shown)


jfilak:
I like this PR.
@sorki what do you think about it?

src/webfaf/templates/reports/item.html
```diff
@@ -33,6 +33,9 @@
         {% else %}
           <h3>Info</h3>
         {% endif %}
+        {% if unpackaged %}
+          Note: Executable of this report does not come from package.
```
xsuchy: from a package

marusak:
@jfilak I pushed new commit.  `ureport: Package element may be missing`.
Let me explain why:
We talked a lot with @mhabrnal  what Satyr and what FAF should accept and what process. We agreed that Satyr should send an empty list if no packages and no DSO list exist. Matej worked on Satyr, but he discovered that Satyr is quite "closely connected" with packages, and without some hacks or rewrite a *large* part of the code it would not work perfectly. Therefore he came to conclusion, that it would be better if `package` element could be missing in uReport. So I did exactly that.

What do you think about it? You were not a huge fan of changing uReport in a first place.

marusak:
Edit: @mhabrnal did succeed with implementing sending empty list when no packages are available.
Now we have a question: is it better to send `packages: []` or omit `packages` element.

jfilak:
> What do you think about it? You were not a huge fan of changing uReport in a first place.

I was just oposed to introducing the IMHO redundant element.

> Now we have a question: is it better to send packages: [] or omit packages element.

I would go for the former, it eplicitly says that there are no packages. The latter looks like a bug to me.

marusak:
> I would go for the former, it eplicitly says that there are no packages. The latter looks like a bug to me.

Agreed. I dropped the new commit and this PR is same as before.

mkutlak:
LGTM!

PR #591, opened 2017-10-02
fix migration - ignore_private_bz

xsuchy:
you cannot add column with nullable=False without default. DB will error that the column contains nullable values


marusak:
Thank you!

PR #593, opened 2017-10-10
docker: Automatic starting of db

marusak:
If run for the first time (docker run), then a new DB is created.
If starting stopped container, then start DB and check for new
migrations.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -33,4 +33,4 @@ VOLUME /var/lib/postgres
 VOLUME /var/spool/faf
 
 EXPOSE 80
-CMD ["sh", "-c", "crond && apachectl -D FOREGROUND"]
+CMD ["sh", "-c", "crond && init_db && apachectl -D FOREGROUND"]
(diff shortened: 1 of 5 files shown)


docker/Dockerfile
```diff
@@ -33,4 +33,4 @@ VOLUME /var/lib/postgres
 VOLUME /var/spool/faf
 
 EXPOSE 80
-CMD ["sh", "-c", "crond && apachectl -D FOREGROUND"]
+CMD ["sh", "-c", "crond && init_db && apachectl -D FOREGROUND"]
```
mkutlak: Wouldn't it be better to create just one init script and include all these into it?

marusak: Great idea! See my new commit in this PR.

docker/files/run_faf
```diff
@@ -27,3 +27,5 @@ else
     sudo -u postgres /usr/bin/pg_ctl start -D /var/lib/postgres/data -w > /dev/null
     sudo -u faf faf-migrate-db
 fi
+crond
+init_db && apachectl -D FOREGROUND
```
mkutlak: There is no init_db anymore, and "-D FOREGROUND" should be "-DFOREGROUND".

marusak: Yeah, I forgot to push the fix with init_db. Fixed. And are you sure about the `-DFOREGROUND`? I think that separated works as well.

mkutlak: Yes, it works, but you get "Passing arguments to httpd using apachectl is no longer supported." warning. -DFOREGROUND is special parameter, check man page of httpd.

mkutlak:
LGTM. Thanks!

PR #597, opened 2017-11-09
kerneloops: Accept oopses without address

marusak:
Due to https://github.com/torvalds/linux/commit/bb5e5ce545f2 now stack
dump does not contain addresses. This is problem, since FAF expects
them. It would be probably possible to get them by ABRT, but from the
commit message it seems like they are not necessary.

I can only see problem while creating problems, since it may be harder
to differentiate one problem from another.

### DIFF
--- a/src/pyfaf/problemtypes/kerneloops.py
+++ b/src/pyfaf/problemtypes/kerneloops.py
@@ -105,7 +105,7 @@ class KerneloopsProblem(ProblemType):
                                   mandatory=False),
 
         "frames":      ListChecker(DictChecker({
-            "address":         IntChecker(minval=0, maxval=((1 << 64) - 1)),
+            "address":         IntChecker(minval=0, maxval=((1 << 64) - 1), mandatory=False),
             "reliable":        Checker(bool),
             "function_name":   StringChecker(pattern=r"^[a-zA-Z0-9_\.]+$",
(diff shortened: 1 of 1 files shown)


marusak:
@jfilak @sorki  What are your options for omitting these values?

jfilak:
I am not sure it would be a wise to try to get the address by ABRT. Sometimes, there are flods of kernel oopses and ABRT chruncing them and trying to identify the corresponding address would make the situation even worse.

Ask your colleagues in charge of testing Kernel - Jan Stancek - and maybe you can have a word with Denys Vlasenko - a former ABRT developer and CC of the linked commit ;).

juliusmilan:
@marusak Regarding "I can only see problem while creating problems, since it may be harder
to differentiate one problem from another."
IMO it may rarely happen that 2 different problems won't be distinguished thanks to avoiding function address, since function addresses are in text segment and changes every time when something is changed in the source code, even unrelated. IMO this concerns just problems inside same function, with same stacktrace.
On the other hand I think it will solve the opposite problem, i.e. when two problems with equal stacktrace wasn't recognized as one due to some small unrelated changes in the source file, which shifted function addresses in whole text segment.

juliusmilan:
as we discussed, my review comment is not really an issue
lgtm

Issue #598, opened 2017-11-10
URGENT! improve link ordering of External URLs


vbenes:
External URLs are super mess.

it's absolutely insane to look for something in external links.
https://url.corp.redhat.com/ad15b94

Would it be possible to sort it time wise? I would prefer to see the newest at the top. So time wise in reverse order. Maybe some sort of separator may be handy too (to separate single report)

Thanks,
Vladimir

marusak:
> Would it be possible to sort it time wise? I would prefer to see the newest at the top. So time wise in reverse order.

Sure, this is one-line fix, since we already save the date, see:
https://github.com/abrt/faf/blob/master/src/pyfaf/storage/report.py#L579
I'll do it ASAP.

> Maybe some sort of separator may be handy too (to separate single report)

Can you elaborate on this more?

marusak:
@vbenes I open PR #600 , which fixes this issue. I hope I understood what you were requesting. Please see the following screenshot. Does it work as you expected? Let me know what would you changed, added or what was misunderstood.

![urls](https://user-images.githubusercontent.com/12330670/32949069-bd0ca490-cba1-11e7-94a0-a0fc96c2a7a0.png)

vbenes:
It looks good. Will the newest report links be on top or the other way around?

marusak:
The newest will be on top. You can see it in the image, The URLS have been send in order first, second, third... so seventh is the newest one.

vbenes:
perfect! that's preferred ordering indeed :-)

xsuchy:
Please add there (i) with bubble which state: URLs are grouped by reports.

PR #600, opened 2017-11-17
Order external urls

marusak:
Fixes abrt/faf#598

### DIFF
--- a/src/webfaf/templates/_helpers.html
+++ b/src/webfaf/templates/_helpers.html
@@ -371,11 +371,14 @@
 {%- endmacro %}
 
 {% macro external_urls(urls) -%}
-{% for url_obj in urls %}
-  <li><a
-    href="{{ url_obj.url }}">{{ url_obj.url }}
-    </a>
-  </li>
+{% for report_url in urls %}
+    {% for url_obj in report_url %}
+      <li><a
+        href="{{ url_obj.url }}">{{ url_obj.url }}
+        </a>
+      </li>
+    {% endfor %}
+    <li class="empty-line"><hr></li>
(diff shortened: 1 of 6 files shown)


src/webfaf/templates/_helpers.html
```diff
@@ -371,11 +371,14 @@
...
-    </a>
-  </li>
+{% for report_url in urls %}
+    {% for url_obj in report_url %}
+      <li><a
+        href="{{ url_obj.url }}">{{ url_obj.url }}
+        </a>
+      </li>
+    {% endfor %}
+    <li class="empty-line"></li>
```
mkutlak: Maybe do something like this:
\<li class="empty-line"\> \<hr /\>  \</li\>

li.empty-line hr {margin: 0.5em 0;} 
li.empty-line:last-child { display: none; }

marusak: hmm, this adds line. It stretches across the whole available space. I don't this it looks good. But I let it for others to decide. Ideas anyone?
after

mkutlak: Yeah, it looks weird when the URLs are short. This is how it looks like with normal URLs. fafpr

marusak:
Updated. The line is now 50% of the width. And I also added tooltips. 
report
problem_full

mkutlak:
The tooltip could say something like this:
"The URLs are grouped by reports. Newest URL is always at the top."
Nothing else comes to my mind.

Otherwise it looks good and can be merged.

marusak:
Updated.

mkutlak:
LGTM. Thank you!

PR #607, opened 2018-03-14
storage: Remove unused imports

marusak:
One unused import after some changes was making tests to fail.

Looked into other files and removed all unused imports in storage
(mainly migrations)


mkutlak:
LGTM.

Issue #588, opened 2017-09-17
Port fedora plugin to new pkgdb


marusak:
https://admin.fedoraproject.org/pkgdb/ was replaced by https://src.fedoraproject.org/ . That means that src/pyfaf/opsys/fedora.py is quite outdated.
Needs to adopt to pagure API to be able to pull releases, components ...

PR #610, opened 2018-03-18
No pkgdb

marusak:
Fixes abrt/faf#588

### DIFF
--- a/src/pyfaf/actions/pull_associates.py
+++ b/src/pyfaf/actions/pull_associates.py
@@ -20,9 +20,9 @@ from pyfaf.actions import Action
 from pyfaf.opsys import systems
 from pyfaf.queries import (get_associate_by_name,
                            get_opsys_by_name,
-                           get_releases)
+                           get_components_by_opsys)
 from pyfaf.storage import (AssociatePeople,
-                           OpSysReleaseComponentAssociate)
+                           OpSysComponentAssociate)
 
 
 class PullAssociates(Action):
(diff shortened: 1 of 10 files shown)


src/pyfaf/actions/pull_associates.py
```diff
@@ -33,66 +33,42 @@ def __init__(self):
...
+            cmdline.opsys = systems.keys()
+
+        opsyss = []
+        for shortname in cmdline.opsys:
             if shortname not in systems:
-                self.log_error("Operating system '{0}' is not installed"
-                               .format(shortname))
-                return 1
+                self.log_warn("Operating system '{0}' is not installed"
+                              .format(shortname))
```
xsuchy: while in this, can you change it to lazy evaluation? `log_warn("foo {0}", shortname)`

src/pyfaf/opsys/fedora.py
```diff
@@ -220,74 +218,68 @@ def save_ureport(self, db, db_report, ureport, packages, flush=False, count=1):
 
     def get_releases(self):
         result = {}
-        collections = self._pkgdb.get_collections()["collections"]
+        url = "https://pdc.fedoraproject.org/rest_api/v1/releases/"
```
xsuchy: I would love to see this as config option.

src/pyfaf/opsys/fedora.py
```diff
@@ -220,74 +218,68 @@ def save_ureport(self, db, db_report, ureport, packages, flush=False, count=1):
...
+        url += "?page_size=-1"
```
xsuchy: What means -1 ?

src/pyfaf/opsys/fedora.py
```diff
@@ -220,74 +218,68 @@ def save_ureport(self, db, db_report, ureport, packages, flush=False, count=1):
...
 
-    def get_component_acls(self, component, release=None):
-        branch = None
-        if release:
-            branch = self._release_to_pkgdb_branch(release)
+        return result
 
+    def get_component_acls(self, component):
         result = {}
+        url = "https://src.fedoraproject.org/api/0/rpms/"
```
xsuchy: config option as well, please

src/pyfaf/storage/migrations/versions/9301a426f19d_associates_to_opsys.py
```diff
@@ -0,0 +1,72 @@
...
+        op.execute(q)
+    """
+
+    # deliberately do not remove OpSysReleaseComponentAssociate as there is
+    # no way how to get those data back.
+
+
+def downgrade():
+    postgresql.ENUM(name="_permission_type").drop(op.get_bind(), checkfirst=False)
+    op.drop_table('opsyscomponentsassociates')
```
xsuchy: should this be in reverse? First drop table and then drop enum?

src/pyfaf/actions/pull_associates.py
marusak: You mean `log_warn("foo %s", shortname)`? If so, I can change it, but it is `the old style` and I am not sure, if the laziness there makes that much of a difference.

src/pyfaf/opsys/fedora.py
marusak: good point. Would it be better to put it in config as pdc_release_url or pdc_url? (It will pull releases and components from pdc, if ti should be two urls in config or one and add to it in code)

src/pyfaf/opsys/fedora.py
marusak: https://github.com/product-definition-center/product-definition-center/issues/504 :)

marusak: It means that all of the results are on one page and not 20 items per page. With releases it is safe as there are around 30 items. With components it is more complicated. There is 18000 of them and for 18000 is 900 requests which is sooo slow.  However I pull there only names, so it is 18000 * len(name) so it is just fine.

marusak:
@xsuchy updated

PR #612, opened 2018-03-20
WIP webfaf: Limit the number of exception mails

mkutlak:
Limits the rate at which the messages are passed to the SMTP handler.
It uses a token bucket based algorithm to filter the messages.

* Ex.1: rate=1 timeframe=30 burst=1
After a mail is sent there is a 30s period until another one can be sent.

* Ex.2: rate=1 timeframe=60 burst=5
Allows to send 5 mails in burst before the filtering kicks in. After that
the number of mails that can be sent starts recovering at the rate of 1 mail
per 60s.

There is also possibility to filter the mails based on the substring
in the exception message.

More info at https://github.com/wkeeling/ratelimitingfilter

### DIFF
--- a/config/plugins/web.conf
+++ b/config/plugins/web.conf
@@ -44,3 +44,11 @@ type = simple
 # memcached_host = localhost
 # memcached_port = 11211
 # memcached_key_prefix = webfaf
+
+[throttle]
+# Number of mails that can be sent per the time frame
+# rate = 1
+# Time to wait before another mail can be sent
+# timeframe = 30
+# Start the throttling after this number of mails is sent
+# burst = 1
--- a/faf.spec.in
+++ b/faf.spec.in
@@ -82,6 +82,7 @@ Requires: python-openid-teams
 Requires: python-jinja2 >= 2.7
 Requires: python-bunch
 Requires: python-dateutil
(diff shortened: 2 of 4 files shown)


xsuchy:
Nice. Lets wait with the merge till the package gets into Fedora stable.

marusak:
What package? python2-ratelimitingfilter ?
Can you post here the build?

mkutlak:
@marusak ```dnf copr enable mkutlak/ratelimitingfilter```

mkutlak:
I've build the ratelimitingfilter package for epel7. It is available in the faf-el7-devel copr repo.

```dnf copr enable @abrt/faf-el7-devel```

marusak:
I think that if we build it in faf-el7 copr, then we are good to go. What do you think @xsuchy ?

xsuchy:
I think it is fine regarding EPEL, but it is not fine for Fedora. I would wait untill this errata hit stable:
  https://bodhi.fedoraproject.org/updates/FEDORA-2018-b22314065a

mkutlak:
I've updated the commit message and commented out the config options.

xsuchy:
python-ratelimitingfilter is now in Fedora, so we are fine to merge

PR #615, opened 2018-03-25
Nonexiting ignore

marusak:


mkutlak:
Couldn't this be fixed in the ```component_names_to_ids``` function?
https://github.com/abrt/faf/blob/2fec071dfe7342b360ba1993b58260ab1b90ba3e/src/webfaf/forms.py#L94
It is also not working in the summary page.
https://github.com/abrt/faf/blob/2fec071dfe7342b360ba1993b58260ab1b90ba3e/src/webfaf/summary.py#L28

PR #617, opened 2018-04-04
py3 compatibility: classic division

ari3s:
The PR supports the same "classic" division in both Python 2 & 3 simultaneously according to [Porting Python 2 Code to Python 3](https://docs.python.org/3/howto/pyporting.html#division) documentation, using python-modernize tool.

### DIFF
--- a/src/pyfaf/actions/stats.py
+++ b/src/pyfaf/actions/stats.py
@@ -19,6 +19,7 @@
 
 from __future__ import unicode_literals
 
+from __future__ import division
 import datetime
 import collections
 
@@ -120,7 +121,7 @@ class Stats(Action):
         for num, (comp, count, reports) in enumerate(results):
             if reports:
                 out += ("{0}. {1} seen {2} times ({3:.0%} of all reports)\n"
-                        .format(num + 1, comp, count, count / float(total_num)))
+                        .format(num + 1, comp, count, count // float(total_num)))
 
                 for problem_url, bugs in reports:
                     if problem_url or bugs:
@@ -182,8 +183,8 @@ class Stats(Action):
                 yysum += y * y
 
             # y = bx + a
-            b = xysum - xsum * ysum / num_days
-            b /= xxsum - xsum ** 2 / num_days
+            b = xysum - xsum * ysum // num_days
+            b /= xxsum - xsum ** 2 // num_days
 
             a = ysum - b * xsum
             a /= num_days
@@ -239,13 +240,13 @@ class Stats(Action):
 
             minval = min(counts)
             maxval = max(counts)
-            scale = ((maxval - minval) << 8) / (len(self.graph_symbols) - 1)
+            scale = ((maxval - minval) << 8) // (len(self.graph_symbols) - 1)
             scale = max(scale, 1)
             graph = ""
 
             for day in prev_days(num_days):
                 graph += self.graph_symbols[((comp.history[day] - minval)
-                                             << 8) / scale]
+                                             << 8) // scale]
 
             out += row.format(component=comp.name,
                               jump=comp.jump,
--- a/src/pyfaf/queries.py
+++ b/src/pyfaf/queries.py
@@ -16,6 +16,7 @@
 # You should have received a copy of the GNU General Public License
 # along with faf.  If not, see <http://www.gnu.org/licenses/>.
 
+from __future__ import division
 from datetime import timedelta
 import datetime
 import functools
@@ -742,7 +743,7 @@ def prioritize_longterm_problems(min_fa, problem_tuples):
         if problem.first_occurrence.day != 1:
             months -= 1
 
-        problem.rank = rank / float(months)
+        problem.rank = rank // float(months)
 
     return sorted(problem_tuples, key=lambda (problem, _, __): problem.rank,
                   reverse=True)
--- a/src/pyfaf/storage/fixtures/__init__.py
+++ b/src/pyfaf/storage/fixtures/__init__.py
@@ -16,6 +16,7 @@
 # You should have received a copy of the GNU General Public License
 # along with faf.  If not, see <http://www.gnu.org/licenses/>.
 
+from __future__ import division
 import sys
 import os
 import time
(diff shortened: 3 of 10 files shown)


trapas:

mhabrnal:
test this please

src/pyfaf/actions/stats.py
```diff
@@ -120,7 +121,7 @@ def components(self, cmdline, db, opsys, release):
         for num, (comp, count, reports) in enumerate(results):
             if reports:
                 out += ("{0}. {1} seen {2} times ({3:.0%} of all reports)\n"
-                        .format(num + 1, comp, count, count / float(total_num)))
+                        .format(num + 1, comp, count, count // float(total_num)))
```
xsuchy: I guess this should stay just /. Or you should change the format string.

src/pyfaf/queries.py
```diff
@@ -742,7 +743,7 @@ def prioritize_longterm_problems(min_fa, problem_tuples):
         if problem.first_occurrence.day != 1:
             months -= 1
 
-        problem.rank = rank / float(months)
+        problem.rank = rank // float(months)
```
xsuchy: This should stay just /

src/pyfaf/storage/fixtures/__init__.py
```diff
@@ -451,7 +452,7 @@ def restore_lob_dir(self, url=None, cache=True):
                     bytes_so_far += len(chunk)
                     f.write(chunk)
 
-                    percent = float(bytes_so_far) / total_size
+                    percent = float(bytes_so_far) // total_size
```
xsuchy: This should stay just /

src/webfaf/filters.py
```diff
@@ -56,7 +57,7 @@ def fancydate(value, base_date=None):
 
     if old_date.month == base_date.month and old_date.year == base_date.year:
         # computes a number of calendar weeks (not only 7 days)
-        offset = round((d.days - base_date.isoweekday()) / 7, 0) + 1
+        offset = round((d.days - base_date.isoweekday()) // 7, 0) + 1
```
xsuchy: This needs to be / so the round() works

src/webfaf/problems.py
```diff
@@ -533,7 +534,7 @@ def item(problem_id):
     # are selected which is not a good representation.)
     k = min(len(bt_hashes), 10)
     a = 0
-    d = len(bt_hashes)/float(k)
+    d = len(bt_hashes)//float(k)
```
xsuchy: I guess this should stay / as well.

src/webfaf/reports.py
```diff
@@ -1014,8 +1015,8 @@ def attach():
 
 def get_avg_count(first, last, count):
     diff = last - first
-    r_d = diff.days / 30.4  # avg month size
+    r_d = diff.days // 30.4  # avg month size
```
xsuchy: Just /

src/webfaf/reports.py
```diff
@@ -1014,8 +1015,8 @@ def attach():
...
     if r_d < 1:
         r_d = 1
 
-    return int(round(count / r_d))
+    return int(round(count // r_d))
```
xsuchy: Just / so the round() can work

src/webfaf/stats.py
```diff
@@ -85,14 +86,14 @@ def date_filter(query):
         if not release_sum:
             continue
 
-        percentage = int(release_sum * 100.0 / total)
+        percentage = int(release_sum * 100.0 // total)
```
xsuchy: When you use // then we do not need the int()

src/webfaf/stats.py
```diff
@@ -85,14 +86,14 @@ def date_filter(query):
...
 
         comps = queries.get_report_count_by_component(
             db, release.opsys.name, release.version, history=history)
 
         comp_data = []
         for comp, count in date_filter(comps).all():
-            comp_percentage = int(count * 100.0 / release_sum)
+            comp_percentage = int(count * 100.0 // release_sum)
```
xsuchy: dtto

src/pyfaf/actions/stats.py
ari3s: You are right, it should stay ```/``` because the second operator is ```float```.

src/pyfaf/queries.py
ari3s: You are right, it should stay ```/``` because the second operator is ```float```.

src/pyfaf/storage/fixtures/__init__.py
ari3s: You are right, it should stay ```/``` because the first operator is ```float```.

src/webfaf/problems.py
ari3s: You are right, it should stay ```/``` because the second operator is ```float```.

src/webfaf/reports.py
ari3s: You are right, it should stay ```/``` because the second operator is ```float```.

src/webfaf/reports.py
ari3s: You are right, it should stay ```/``` because the second operator is ```float``` as the result of the previous division operation (or integer value of 1 when the final result is ```count``` value).

src/webfaf/stats.py
ari3s: It should stay ```/``` because the first division operator is ```float```.

src/webfaf/stats.py
ari3s: It should stay ```/``` because the first division operator is ```float```.

src/webfaf/filters.py
ari3s: I think that the first operator is ```int``` or ```long``` and the second one is ```int```, so this commit is right to ensure the same behavior in the both Python 2 and 3 versions.

xsuchy: Yes, you're right.

xsuchy:
closing per request, Jan will create new PR

PR #620, opened 2018-04-10
py3 compatibility: classic division - corrected

ari3s:
Support of the same "classic" division in both Python 2 & 3 simultaneously.
It is related to such a division when both operators are ```int```.

### DIFF
--- a/src/pyfaf/actions/stats.py
+++ b/src/pyfaf/actions/stats.py
@@ -19,6 +19,7 @@
 
 from __future__ import unicode_literals
 
+from __future__ import division
 import datetime
 import collections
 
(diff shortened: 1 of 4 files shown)


src/pyfaf/actions/stats.py
```diff
@@ -182,8 +183,8 @@ def trends(self, cmdline, db, opsys, release):
                 yysum += y * y
 
             # y = bx + a
-            b = xysum - xsum * ysum / num_days
-            b /= xxsum - xsum ** 2 / num_days
+            b = xysum - xsum * ysum // num_days
+            b /= xxsum - xsum ** 2 // num_days
```
mhabrnal: You should change `/=` to `//=` as well.

src/pyfaf/actions/stats.py
```diff
@@ -182,8 +183,8 @@ def trends(self, cmdline, db, opsys, release):
...
 
             a = ysum - b * xsum
             a /= num_days
```
mhabrnal: the same here (`//=` -> `/=`)

mkutlak:
retest this please

PR #630, opened 2018-04-14
py3 compatibility: Replace StringIO and cStringIO modules with io module

ari3s:
The `StringIO` and `cStringIO` modules are gone in Python 3.
Instead, import the `io` module and use `io.StringIO` or `io.BytesIO`
for text and data respectively.

### DIFF
--- a/src/pyfaf/utils/contextmanager.py
+++ b/src/pyfaf/utils/contextmanager.py
@@ -1,5 +1,12 @@
 import sys
-from StringIO import StringIO
+
+if sys.version_info.major == 2:
+#Python 2
+    from StringIO import StringIO
+else:
+#Python 3+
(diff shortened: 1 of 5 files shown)


src/pyfaf/utils/contextmanager.py
```diff
@@ -1,5 +1,12 @@
 import sys
-from StringIO import StringIO
+
+try:
+#Python 2
+    from StringIO import StringIO
+except ImportError:
+#Python 3
+    from io import StringIO
+
```
xsuchy: It is probably matter of preference, but I prefer:

```
if six.PY3:
  ...
```
Your style can hide issues when it is Python2 and there is some error during import of StringIO (other than module is missing).

PR #631, opened 2018-04-17
py3 compatibility: Replacement of zip with six.moves.zip

ari3s:
In Python 2, the global `zip()` function took any number of sequences
and returned a list of tuples. In Python 3, `zip()` returns an iterator
instead of a list.

### DIFF
--- a/src/pyfaf/utils/format.py
+++ b/src/pyfaf/utils/format.py
@@ -17,6 +17,7 @@
 # along with faf.  If not, see <http://www.gnu.org/licenses/>.
 
 from functools import reduce
+from six.moves import zip
 
 __all__ = ["as_table"]
 
@@ -31,7 +32,7 @@ def as_table(headers, data, margin=1, separator=' '):
 
     widths = reduce(
         lambda x, y: map(
-            lambda a_b: max(a_b[0], a_b[1]), zip(x, y)
+            lambda a_b: max(a_b[0], a_b[1]), list(zip(x, y))
         ),
         map(lambda x: map(len, x), data) + [map(len, headers)],
         map(lambda _: 0, headers))


marusak:
Are those absolute imports necessary there?

src/pyfaf/utils/format.py
```diff
@@ -31,7 +33,7 @@ def as_table(headers, data, margin=1, separator=' '):
 
     widths = reduce(
         lambda x, y: map(
-            lambda a_b: max(a_b[0], a_b[1]), zip(x, y)
+            lambda a_b: max(a_b[0], a_b[1]), list(zip(x, y))
```
marusak: I am confused - you imported zip from six.moves so the zip returns list, right? Why are you then calling list on the list?

ari3s: It returns iterator (you can see [https://www.eflglobal.com/version-agnostic-python-part-1/](https://www.eflglobal.com/version-agnostic-python-part-1/)).

src/pyfaf/actions/create_problems.py
```diff
@@ -31,6 +32,7 @@
                            get_unassigned_reports,
                            get_problem_by_id)
 from pyfaf.storage import Problem, ProblemComponent, Report
+from six.moves import zip
```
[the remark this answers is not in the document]
ari3s: You are right, I will return the change back.

src/pyfaf/utils/format.py
xsuchy: six.moves just make sure that the name of the function is the same. Which is proably not needed, because zip() has the same name in both python2 and python3 and it work for me even without the import. But true is that every documentation mention this import.

PR #635, opened 2018-04-19
py3 compatibility: Add list function to dictionary methods

ari3s:
Many dictionary methods return lists in Python 2, but all of these
methods return dynamic views in Python 3. In some context, the problem
is solved by adding the `list()` function.


src/webfaf/problems.py
```diff
@@ -509,11 +509,11 @@ def item(problem_id):
         names[pkg.name]["versions"][pkg.evr()] += cnt
 
     package_counts = []
-    for pkg in sorted(names.values(), key=itemgetter("count"), reverse=True):
+    for pkg in sorted(list(names.values()), key=itemgetter("count"), reverse=True):
```
xsuchy: sorted() explicitly convert result to list, no need for list() here.

src/webfaf/reports.py
```diff
@@ -521,11 +521,11 @@ def item(report_id, want_object=False):
             .format(pkg.iepoch, pkg.iversion, pkg.irelease)] += pkg.count
 
     package_counts = []
-    for pkg in sorted(names.values(), key=itemgetter("count"), reverse=True):
+    for pkg in sorted(list(names.values()), key=itemgetter("count"), reverse=True):
```
xsuchy: and here

src/webfaf/reports.py
```diff
@@ -521,11 +521,11 @@ def item(report_id, want_object=False):
...
         package_counts.append((
             pkg["name"],
             pkg["count"],
-            sorted(pkg["versions"].items(), key=itemgetter(1), reverse=True)))
+            sorted(list(pkg["versions"].items()), key=itemgetter(1), reverse=True)))
```
xsuchy: and here

ari3s:
@xsuchy thank you for the review! I have updated the PR according to your hints.

PR #647, opened 2018-04-28
Clean up of unused and duplicated imports

ari3s:


marusak:
What about  
`src/bin/faf-migrate-db:27` ?

And could you supress warning on `./src/pyfaf/utils/web.py:11` as is done in `./src/pyfaf/actions/shell.py`? (if you use pylint - do you?)

ari3s:
@marusak: Thank you for the review. I have fixed the missing unused import at `src/bin/faf-migrate-db:27`.

Regarding the second point, I use pylint filtering `unused-import` which does not detect this warning. Do you mean to fix it in the way of `./src/pyfaf/actions/shell.py:57..61`?

marusak:
> Regarding the second point, I use pylint filtering `unused-import` which does not detect this warning. 

Weird, it is marked in pylint as `Unused variable`...

```
pylint-3 ./src/pyfaf/utils/web.py
>>>> .....
>>>>W: 12, 8: Unused variable 'webfaf' (unused-variable)
>>>>.....
```
> Do you mean to fix it in the way of `./src/pyfaf/actions/shell.py:57..61`?

No, I meant `./src/pyfaf/actions/shell.py:23..24`, but now I've realized it needs to be like `./src/pyfaf/actions/shell.py:20`.

Anyway, as pylint takes it as different error, you may ignore my comment and this may be fixed in next PR, that will fix unused variables.

PR #651, opened 2018-05-10
py3 compatibility: Adjust imports

ari3s:
In Python 3, imports can be either absolute or explicitly relative;
implicit relative imports are forbidden.


xsuchy:
looks good to me

PR #674, opened 2018-06-05
save-reports: Enable matching reports to save

marusak:
This may be helpful when there is way to many reports to process. `--speedup` takes way too long to lock files and also only one process works on saving.

### DIFF
--- a/src/pyfaf/actions/save_reports.py
+++ b/src/pyfaf/actions/save_reports.py
@@ -146,16 +146,16 @@ class SaveReports(Action):
         db_unknown_opsys.count += 1
         db.session.flush()
 
-    def _save_reports(self, db):
+    def _save_reports(self, db, pattern="*"):
         self.log_info("Saving reports")
 
-        report_filenames = os.listdir(self.dir_report_incoming)
+        report_filenames = glob.glob(os.path.join(self.dir_report_incoming, pattern))
 
         i = 0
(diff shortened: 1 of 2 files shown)


src/pyfaf/actions/save_reports.py
```diff
@@ -372,3 +375,5 @@ def tweak_cmdline_parser(self, parser):
         parser.add_argument("--speedup", action="store_true",
                             default=False, help="Speedup the processing. "
                             "May be less accurate.")
+        parser.add_argument("--pattern", help="Save reports matched with pattern. "
+                            "Does not work with speedup. E.g. --pattern \"123*\"")
```
mkutlak: Why not create a mutual exclusive group then? 

https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group

src/pyfaf/actions/save_reports.py
```diff
@@ -359,7 +359,10 @@ def run(self, cmdline, db):
                     os.remove(self.lock_filename)
                     raise
             else:
```
mkutlak: If the pattern doesn't work with speedup, the code can be changed to.
```
if cmdline.speedup:
   ...
elif cmdline.pattern:
   ...
else:
   ...
```

marusak:
@mkutlak updated

Issue #595, opened 2017-10-19
Does FAF need BZ comments and  attachments?


marusak:
What does it FAF use for? How important is it for FAF? It seems to me, that it only takes up space (around 10GB on retrace.fedoraproject.org - it is not too much, but I am trying to understand what is used for)
@sorki @jfilak  ?

jfilak:
Not sure. Remove them and you will see :)

In the history, ABRT services used to add automatic comments. Those comments could be parsed out and some important information could be gathered from them. Full featured backtrace is an example of such an important information - but that's history.

@mtoman Do you remember why FAF needs BZ comments?

mtoman:
I have two equally probable theories :)
1) They were used to match duplicates, because marking a bug as a duplicate creates a comment which is persistent even if the bug changes state later on and somebody changes/removes the duplicate field.
2) It was sort of feature-completeness thing - the first versions of FAF did not know what data they would use in the future so they cached everything they could.

mtoman:
Actually
3. in very early versions, before uReports, FAF could have been looking for comments from ABRT and use them to reconstruct the metadata that are now sent in uReports

marusak:
Thank you for clarification. Can it be removed? Or should we keep it and only add config option not to save those data?

jfilak:
Remove the comments and attachments (you have several terrabytes of backups), turn of the feature, don't remove the feature from code, good plan?

marusak:
> Remove the comments and attachments (you have several terrabytes of backups), turn of the feature, don't remove the feature from code, good plan?

What do you mean by `turn off`? With config, or remove calling the function, which does this saving? (and keep implemented functions for saving these thinks).

PR #682, opened 2018-06-14
Allow users to see and delete data associated with them

mkutlak:


src/pyfaf/queries.py
```diff
@@ -21,59 +21,8 @@
...
-                           ReportUnknownPackage,
-                           SfPrefilterBacktracePath,
-                           SfPrefilterPackageName,
-                           SfPrefilterSolution,
-                           Symbol,
-                           SymbolSource,
-                           UnknownOpSys)
-
 from pyfaf.opsys import systems
+from pyfaf.storage import *
```
marusak: Why this? It is a bad practice to use wildcard import

src/pyfaf/queries.py
```diff
@@ -1416,3 +1371,123 @@ def get_all_report_hashes(db, date_from=None,
...
+    Delete Bugzilla for given bug_id.
+    """
+    query = (db.session.query(BzBug)
+             .filter(BzBug.duplicate == bug_id)
+             .all())
+
+    for bgz in query:
+        bgz.duplicate = None
+
+    db.session.query(BzComment).filter(BzComment.bug_id == bug_id).delete(False)
```
marusak: would be nice to also delete the lob (if there is one)

src/pyfaf/queries.py
```diff
@@ -1416,3 +1371,123 @@ def get_all_report_hashes(db, date_from=None,
...
+    db.session.query(BzBugCc).filter(BzBugCc.bug_id == bug_id).delete(False)
+    db.session.query(BzBugHistory).filter(BzBugHistory.bug_id == bug_id).delete(False)
+    db.session.query(BzAttachment).filter(BzAttachment.bug_id == bug_id).delete(False)
```
marusak: also the lob

src/webfaf/utils.py
```diff
@@ -374,3 +376,30 @@ def stream_template(template_name, **context):
     rv = t.stream(context)
     rv.enable_buffering(2)
     return rv
+
+
+def delete_user_bugzillas(db, user_id):
```
marusak: wondering about this - do we need to remove bugzilla? Could not we only remove the user? I mean, we could only forget who created the bz, who made what changes (history) etc. but not remove the bz from faf and from report?

src/webfaf/user.py
```diff
@@ -0,0 +1,72 @@
...
+
+    if fas_user is not None:
+        queries.get_reportarchives_by_username(db, fas_user.username).delete(False)
+        queries.get_problemreassigns_by_username(db, fas_user.username).delete(False)
+
+    if contact_email is not None:
+        user_reports = queries.get_reportcontactmails_by_id(db, contact_email.id)
+
+        for report in user_reports.all():
+            delete_complete_report(db, report.report_id)
```
marusak: soo, you remove the report, if the user attached his/hers email to it? This is soo unsafe, because I can attach my email to any existing report (for example look at `reporter-ureport --help` how easy is to do that) and then remove any report. This way I can clear our db easily :)

marusak: Now I saw, that in the given function there is the check if the count is more than one, little bit confusing to call method `delete_complete_report` but delete only when the count is less than 2 :) Still there is the problem I can remove selected report, but sure, only if the count is 1.

src/webfaf/utils.py
mkutlak: How about removing the user, his attachments and comments (these are not needed? https://github.com/abrt/faf/issues/595) and also the ccs (useless information without the user).
And anonymize the bugzilla and the history.

Does that make sense?

src/pyfaf/queries.py
mkutlak: Will do.

src/webfaf/utils.py
marusak: yeah, that makes more sense. I just don't want to remove the bugzzilla ID and resolution, as it is rather usefull. And just showing, that bugzilla 123 is related to problem 456 is not showing any personal info.

src/webfaf/user.py
mkutlak: If the `count` is 1 and there is a `contact mail` linked to it then the reports comes from the user, right?



> little bit confusing to call method delete_complete_report but delete only when the count is less than 2 :)

I will change that.

marusak: not necessary. Under normal circumstances it is true, but anyone can attach its email to any report. So I can pick any report on faf with count==1, use reporter-ureport to attach my email to it and then remove it.

mkutlak:
Updated, I will add tests later.

src/pyfaf/queries.py
```diff
@@ -1389,30 +1344,150 @@ def get_all_report_hashes(db, date_from=None,
...
+    query(st.ReportReason).filter(st.ReportReason.report_id == report_id).delete(False)
+    query(st.ReportReleaseDesktop).filter(st.ReportReleaseDesktop.report_id == report_id).delete(False)
+    query(st.ReportSelinuxContext).filter(st.ReportSelinuxContext.report_id == report_id).delete(False)
+    query(st.ReportSelinuxMode).filter(st.ReportSelinuxMode.report_id == report_id).delete(False)
+    query(st.ReportUnknownPackage).filter(st.ReportUnknownPackage.report_id == report_id).delete(False)
+    query(st.ReportUptime).filter(st.ReportUptime.report_id == report_id).delete(False)
+    query(st.ReportURL).filter(st.ReportURL.report_id == report_id).delete(False)
+
+    # delete backtraces
+    bt = query(st.ReportBacktrace).filter(st.ReportBacktrace.report_id == report_id).first()
```
marusak: Are we sure, that there will be always one bt? If there will be none, the following line will fail (None.id) and if there are more than one, only one bt will be deleted. The same with thread.

src/webfaf/utils.py
```diff
@@ -374,3 +377,50 @@ def stream_template(template_name, **context):
...
+
+
+def delete_bugzilla_user(db, user_id, alt_id):
+    """
+    For given user_id delete BzUser and his comments, attachments, ccs from the database.
+    And replace 'user_id' in related bugzillas and bugzilla history with 'alt_id'.
+    """
+    bzcomments = queries.get_bzcomments_by_uid(db, user_id).all()
+    for bzcomm in bzcomments:
+        if bzcomm.has_log("content"):
```
marusak: shouldn't this be `has_lob`?

mkutlak:
Updated: corrected the typo, and removed the delete report function from queries.

marusak:
retest this please

PR #684, opened 2018-06-20
pylint: Fix inconsistent return statements

marusak:

### DIFF
--- a/src/webfaf/problems.py
+++ b/src/webfaf/problems.py
@@ -639,4 +639,4 @@ def bthash_forward(bthash=None):
         return render_template("problems/multiple_bthashes.html",
                                problems=problems)
 
-    abort(404)
+    return abort(404)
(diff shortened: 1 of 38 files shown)


mkutlak:
retest this please

src/webfaf/problems.py
```diff
@@ -640,3 +640,4 @@ def bthash_forward(bthash=None):
                                problems=problems)
 
     abort(404)
+    return None
```
mkutlak: The `return None` will never be executed because the `abort(404)` will redirect you to an error page before it.

marusak: I know, but pylint doesn't. There are two options - this return or disable pylint warning on this function and I found this less harmful

mkutlak: You can change it to `return abort(404)`, that should stop pylint complaining.

marusak:
@mkutlak updated

src/webfaf/reports.py
```diff
@@ -347,6 +347,7 @@ def get_hash(os=None, release=None, since=None, to=None):
         return jsonify({"data": r_hash})
 
     abort(405)
+    return None
```
mkutlak: `return abort(405)`

marusak:
@mkutlak updated

mkutlak:
Thank you

PR #685, opened 2018-06-21
migrations: Statement must be first

marusak:
If the statement is not before import, then the explaining text is not
printed while migrating.


mkutlak:
Thank you

PR #699, opened 2018-08-06
webfaf: Unbundle patternfly and other assets

mkutlak:
Unbundle [Patternfly](https://github.com/patternfly/patternfly) and put custom css modifications for FAF to separate file.
Part of the PR are fixes for bugs that somehow broke the style of the web.

Other unbundled assets:
- jQuery
- DataTables
- moment.js
- typeahead.js

Container:
`docker pull mkutlak/faf-unbundle`

### DIFF
--- a/src/webfaf/static/css/style.css
+++ b/src/webfaf/static/css/style.css
@@ -1124,10 +1124,19 @@ fieldset[disabled] .datepicker table tr td span.active.disabled:hover.focus {
   .navbar-pf .navbar-utility > li > a:hover,
   .navbar-pf .navbar-utility > li > button:hover {
     border-left-color: #6c95cf;
+    background: rgba(255, 255, 255, 0.14);
   }
   .navbar-pf .navbar-utility > li.open > a,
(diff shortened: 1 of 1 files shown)


marusak:
> Some of the packages are not available for EL7 yet.

What does mean `yet`?

marusak:
Tables are overflowing. Note that in the current code when zooming too much, tables do not fit the page. In this code, tables always fit the page (which I like more, to be honest), but I think it is unacceptable that some fields may overflow (can be found on more places).
b1-overflow
overflow

mkutlak:
> Tables are overflowing.

This should be fixed with the last commit.


> Note that in the current code when zooming too much, tables do not fit the page. In this code, tables always fit the page

The last commit sort of goes back to the current behaviour - if the page width is less than ~650px (width of reports/problems table header) then the tables won't fit the page.

mkutlak:
All the required packages should now be available for EL (from Copr repo) and Fedora. 

And I've corrected colors for menu in navigation bar.

PR #706, opened 2018-09-14
tell logrotate to rotate files under faf:faf

xsuchy:
otherwise it emits warning and will not logrotate the files


mhabrnal:
Excellent work!!!

PR #720, opened 2018-10-26
Python3 completed

marusak:
Closes #702 

This is spin-off from #702 . I've found some issues, fixed those. Also I've found out why the test was not working and fixed that as well. Moreover I've cleaned up commits a bit, so I've decided to open a new PR.

Credit however goes to @ari3s as main author

### DIFF
--- a/src/webfaf/blueprints/symbol_transfer.py
+++ b/src/webfaf/blueprints/symbol_transfer.py
@@ -37,8 +37,8 @@ def process_symbol(build_id, path, offset, problem_type, create_symbol_auth_key)
             # We need to attach our symbols to a dummy report in order to set
             # their type
             h = sha1()
-            h.update("symbol_transfer_dummy")
-            h.update(problem_type)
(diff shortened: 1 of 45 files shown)


src/webfaf/blueprints/symbol_transfer.py
```diff
@@ -39,7 +39,7 @@ def process_symbol(build_id, path, offset, problem_type, create_symbol_auth_key)
             h = sha1()
             h.update("symbol_transfer_dummy")
             h.update(problem_type)
-            dummy_report_hash = h.hexdigest()
+            dummy_report_hash = h.encode("utf-8").hexdigest()
```
mkutlak: webfaf/blueprints/symbol_transfer.py:42:32: E1101: Instance of 'sha1' has no 'encode' member (no-member)

marusak: Nice catch - fixed :)

PR #731, opened 2018-11-08
Finish pylint

marusak:
These commits finish 10/10 pylint checks.
It works on Fedora 29 local machine. It does not pass either in copr or in docker or on python2 based systems.
These issues will follow up (not building in copr is because of some missing buildrequires, docker is now with F27 and python2 build does not bring python3 packages, which pylint3 requires - maybe we should use pylint2 for python2 builds?)


xsuchy:
+1

PR #733, opened 2018-11-14
Define message schema and move to fedora-messaging from fedmsg

mkutlak:
Adds `faf-schema` subpackage that defines message schema for the `fedora-messaging`
infrastructure. The `fedora-messaging` library is designed to be a replacement for
the `fedmsg` library.

Message schema should be treated as immutable. Once defined, they should not be altered. 
Instead, define a new schema class, mark the old one as deprecated, and remove it after an appropriate transition period.[1]

Docs:
[1] fedora-messaging.readthedocs.io/en/latest/messages.html#schema-are-immutable
fedora-messaging.readthedocs.io/en/latest/publishing.html
fedora-messaging.readthedocs.io/en/latest/tutorial/conversion.html

### DIFF
--- a/src/schema/faf_schema/schema.py
+++ b/src/schema/faf_schema/schema.py
@@ -1,0 +1,132 @@
+# Copyright (C) 2018  Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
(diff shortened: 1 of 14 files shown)


src/schema/faf_schema/schema.py
```diff
@@ -0,0 +1,132 @@
...
+        return "Property 'occurance' not implemented in the FafMessage."
+
+
+class FafProblemMessage(FafMessage):
+    """
+    A sub-class of a Fedora message that defines a message schema for
+    messages published by FAF to notify about a problem.
+    """
+    body_schema = {
+        "id": "http://fedoraproject.org/message-schema/faf#",
```
xsuchy: Should this be a valid URL?

xsuchy:
+1 once merged, we should not forget to correctly configure (especially dev) instance according to https://fedora-messaging.readthedocs.io/en/latest/configuration.html#config

src/schema/faf_schema/schema.py
mkutlak: I am not sure. It is from the [fedora-messaging#creating-the-schema-package](https://fedora-messaging.readthedocs.io/en/latest/tutorial/schemas.html#creating-the-schema-package)
```id (or $id): an URI identifying this schema. Change the last part of the example URL to use your app’s name.```

Same in [#converting-a-flask-app](https://fedora-messaging.readthedocs.io/en/latest/tutorial/conversion.html).

So maybe in the future.

PR #737, opened 2018-12-06
defer loading pickle of known pulled reports in pull_reports

xsuchy:
__init__() of the action is run always - even when you do not run the action itself, but some other
since load_known() loads pickle, which can be 100 MB, the memory structure can grow to up to 4 GB, which is wasting of
memory, when we actually do not need the data.

### DIFF
--- a/src/pyfaf/actions/pull_reports.py
+++ b/src/pyfaf/actions/pull_reports.py
@@ -49,8 +49,6 @@ class PullReports(Action):
 
         self.known_file = os.path.join(self.basedir, "reports",
                                        PullReports.KNOWN_FILE_NAME)
-        self._load_known()
-
         self.incoming_dir = os.path.join(self.basedir, "reports", "incoming")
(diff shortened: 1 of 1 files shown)


src/pyfaf/actions/pull_reports.py
```diff
@@ -140,6 +138,8 @@ def _get_report(self, name):
     def run(self, cmdline, db):
         if cmdline.master is not None:
             self.master = cmdline.master
+        # can come from cmdline or from config
+        if self.master:
```
marusak: This is always true.

src/pyfaf/actions/pull_reports.py
```diff
@@ -140,6 +138,8 @@ def _get_report(self, name):
...
             self._load_known()
```
marusak: We just need to move line 143 one level up, as we want to load known on each run.

PR #740, opened 2018-12-13
Enable automatic build in copr

marusak:


xsuchy:
This is just target for SRPM. It does nothing with Copr directly (as subject suggest). Merging thou.

PR #749, opened 2019-01-21
WIP: Drop support for Python 2 and yum

ernestask:
One thing I’m not sure about is what to do with the repo table of the database. Add a migration script to replace the type with “dnf”?


autogen.sh
```diff
@@ -6,12 +6,11 @@ cat << EOH
 Prepares the source tree for configuration
 
 Usage:
-  autogen.sh [sydeps [--install-yum|--install-dnf]]
+  autogen.sh [sydeps [--install-dnf]]
```
marusak: it would be better to have only --install (as we have in abrt, libreport...)

src/bin/faf-migrate-db
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
```
xsuchy: In Fedora, we prefer /usr/bin/python3 in shebang.

ernestask: Changed it. Looks like we’ve been sedding these out in the spec file, but I just ripped that out and modified the files directly.

xsuchy:
Lets do the migration in separate PR and merge this one.

PR #751, opened 2019-01-24
retrace: Decode output of popen

mkutlak:
Strings were printed out as hex characters (ie. f = \x66).

### DIFF
--- a/src/pyfaf/retrace.py
+++ b/src/pyfaf/retrace.py
@@ -258,7 +258,7 @@ def demangle(mangled):
     if child is None:
         return None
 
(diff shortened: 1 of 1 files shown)


ernestask:
In Python 3, an alternative could be specifying the encoding in the `subprocess.Popen` constructor:
https://github.com/abrt/faf/blob/ce87541823cbb59873ab59ea9ea24d3c595b4bec/src/pyfaf/utils/proc.py#L35

https://docs.python.org/3/library/subprocess.html#subprocess.Popen.stdout

Looks good to me otherwise.

PR #756, opened 2019-02-12
core: Check if unpacked binary path is defined

mkutlak:
I don't know what can cause this but the check should prevent from errors
when the unpacked_path variable stays None.


mkutlak:
@trapas retest this please

PR #760, opened 2019-02-28
fedmsg: Convert components names to list

mkutlak:
Validator from faf messaging schema requires component names to be a list of string values.

https://github.com/abrt/faf/blob/8b9a3444273fcc4476eec01baf808098c3637084/src/schema/faf_schema/schema.py#L58


ernestask:
https://fedora-messaging.readthedocs.io/en/latest/tutorial/schemas.html#schema-format for reference.

Issue #764, opened 2019-03-18
Tests failing with SQLAlchemy 1.3+


ernestask:
These ones in particular, but the cause seems to be the same:

  * test_problems.py
  * test_reports.py
  * test_summary.py

```
======================================================================
ERROR: test_problem_shown (__main__.ProblemsTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/engine/base.py", line 1179, in _execute_context
    context = constructor(dialect, self, conn, *args)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/engine/default.py", line 683, in _init_compiled
    for grp, m in enumerate(parameters)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/engine/default.py", line 683, in <listcomp>
    for grp, m in enumerate(parameters)
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/sql/compiler.py", line 692, in construct_params
    pd[self.bind_names[bindparam]] = bindparam.effective_value
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/sql/elements.py", line 1213, in effective_value
    return self.callable()
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/orm/relationships.py", line 1580, in _go
    % (column, state_str(state))
sqlalchemy.exc.InvalidRequestError: Can't resolve value for column reports.id on object <Report at 0x7f08ed553f60>; no value has been set for this column

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "./tests/test_webfaf/test_problems.py", line 36, in setUp
    self.save_report("ureport_kerneloops")
  File "/home/ekulik/Development/github.com/abrt/faf/tests/faftests/__init__.py", line 229, in save_report
    self.save_report_dict(self.load_report(filename))
  File "/home/ekulik/Development/github.com/abrt/faf/tests/faftests/__init__.py", line 211, in save_report_dict
    ureport.save(self.db, report, timestamp=mtime)
  File "/home/ekulik/Development/github.com/abrt/faf/src/pyfaf/ureport.py", line 345, in save
    timestamp=timestamp, count=count)
  File "/home/ekulik/Development/github.com/abrt/faf/src/pyfaf/ureport.py", line 235, in save_ureport2
    db_reportosrelease = get_reportosrelease(db, db_report, db_osrelease)
[...]

PR #767, opened 2019-03-26
Fix post-SQLAlchemy-1.3 breakage

ernestask:
I’m fairly sure that the missing primary keys wouldn’t be a problem if we had [`autoflush`](https://docs.sqlalchemy.org/en/latest/orm/session_api.html#sqlalchemy.orm.session.Session.params.autoflush) enabled (and [`autocommit`](https://docs.sqlalchemy.org/en/latest/orm/session_transaction.html#autocommit-mode) is considered legacy, so maybe would be beneficial to look into it deeper).

Closes https://github.com/abrt/faf/issues/764

### DIFF
--- a/src/webfaf/problems.py
+++ b/src/webfaf/problems.py
@@ -65,8 +65,8 @@ def query_problems(_, hist_table,
 
     rank_query = (db.session.query(Problem.id.label('id'),
                                    func.sum(hist_table.count).label('rank'))
-                  .join(Report)
-                  .join(hist_table))
+                  .join(Report, Report.problem_id == Problem.id)
+                  .join(hist_table, hist_table.report_id == Report.id))
     if opsysrelease_ids:
         rank_query = rank_query.filter(
             hist_table.opsysrelease_id.in_(opsysrelease_ids))
--- a/src/pyfaf/actions/create_problems.py
+++ b/src/pyfaf/actions/create_problems.py
@@ -217,6 +217,7 @@ class CreateProblems(Action):
                     # No possible match found, must be a new problem
                     db_problem = Problem()
                     db.session.add(db_problem)
+                    db.session.flush()
                     created_count += 1
                 else:
                     # Leave the problems for the second phase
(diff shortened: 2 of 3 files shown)


ernestask:
Manual flushing is done pretty much everywhere, so seems like this is the way to go for now.

mkutlak:
Reference for the 1.3 changes:
https://docs.sqlalchemy.org/en/latest/changelog/migration_13.html#query-join-handles-ambiguity-in-deciding-the-left-side-more-explicitly

mkutlak:
I think adding flushes everywhere is not the way to go.

Instead, in queries where we use DB objects we should check if the queried object has PK assigned. 
See:
https://github.com/mkutlak/faf/commit/5266c7770303ebc22ffbd2b8827ea27d10ea9056

Why?:
After we've just created a new `Report` it doesn't make sense to use it as a filter to check what `OsRelease` it has assigned; it will return `None` anyway.

See:
https://github.com/abrt/faf/blob/ace9c7080a64076b0938cbb0d135825da74e643c/src/pyfaf/ureport.py#L235


Reference:
https://docs.sqlalchemy.org/en/latest/changelog/migration_10.html#warnings-emitted-when-comparing-objects-with-none-values-to-relationships

src/webfaf/problems.py
```diff
@@ -65,8 +65,8 @@ def query_problems(_, hist_table,
 
     rank_query = (db.session.query(Problem.id.label('id'),
                                    func.sum(hist_table.count).label('rank'))
-                  .join(Report)
-                  .join(hist_table))
+                  .join(Report, Report.problem_id == Problem.id)
+                  .join(hist_table, hist_table.report_id == Report.id))
```
mkutlak: Just a nitpick.. you don't need to specify the `... ON ...` for the `hist_table`. 
You only need to explicitly specify the first join (because it's not clear whether `Problem.id.label('id')` should be used or the `func.sum(hist_table.count).label('rank')`).

ernestask: I’m leaning towards being (overly) explicit, even if there won’t be more columns to confuse things in the near future.

PR #777, opened 2019-04-10
docker: Replace apachectl with httpd to start the server

mkutlak:
With new version of httpd `apachectl` no longer accepts the `-DFOREGROUND` option.
`apachectl: The "-DFOREGROUND" option is not supported.`


ernestask:
But `apachectl` should only pass the option through: https://github.com/apache/httpd/blob/trunk/support/apachectl.in.

ernestask:
Oh, it’s probably the quotes.

PR #781, opened 2019-04-18
pyfaf: repos: rpm_metadata: Flatten URL argument

ernestask:
As the constructor collects all remaining positional arguments as URLs,
we end up receiving a tuple (usually 1-tuple) with a list as the first
element. In places where the URLs are iterated as a list, this will
quickly break things that expect strings and not lists.

### DIFF
--- a/src/pyfaf/repos/rpm_metadata.py
+++ b/src/pyfaf/repos/rpm_metadata.py
@@ -122,7 +122,7 @@ class RpmMetadata(Repo):
 
     name = "rpmmetadata"
 
-    def __init__(self, name, *urls):
+    def __init__(self, name, urls, *args):
         """
         Following `url` schemes are supported:
         http://, ftp://, file:// (used if full
--- a/tests/test_rpm_metadata.py
+++ b/tests/test_rpm_metadata.py
@@ -118,7 +118,7 @@ class RpmMetadataTestCase(faftests.TestCase):
         repository without any protocol in URL correctly.
         """
 
-        rpm_metadata = RpmMetadata("test_repo_absolute", self.tmpdir)
+        rpm_metadata = RpmMetadata("test_repo_absolute", [self.tmpdir])
         rpm_metadata.cachedir = self.cachedir
 
         pkgs = rpm_metadata.list_packages(["noarch"])
@@ -133,7 +133,7 @@ class RpmMetadataTestCase(faftests.TestCase):
         repository with file:// protocol in URL correctly.
         """
 
-        rpm_metadata = RpmMetadata("test_repo_file", "file://" + self.tmpdir)
+        rpm_metadata = RpmMetadata("test_repo_file", ["file://" + self.tmpdir])
         rpm_metadata.cachedir = self.cachedir
 
         pkgs = rpm_metadata.list_packages(["noarch"])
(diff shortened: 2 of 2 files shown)


ernestask:
More or less a copy of https://github.com/abrt/faf/blob/f03bf94099b172fbbc300e14d233fc1a4f833720/src/pyfaf/repos/dnf.py#L58.

mkutlak:
Is this really needed? `*urls` is a tuple of urls. The only place where the `urls` are used is this one:
https://github.com/abrt/faf/blob/f796e816b2e93bb9af75d4f454d6355505eef7c3/src/pyfaf/repos/rpm_metadata.py#L260
___
I don't see a way to get a list of urls to the `*urls` tuple like this: `urls = ([ftp:///a, file:///b], ftp:///c)`

src/pyfaf/repos/rpm_metadata.py
```diff
@@ -143,7 +143,13 @@ def __init__(self, name, *urls):
...
 
         self.name = name
-        self.urls = urls
+        self.urls = []
+        for url in urls:
+            if isinstance(url, str):
+                self.urls.append(url)
+            else:
+                for url_single in url:
+                    self.urls.append(url_single)
```
xsuchy: these two lines can be replaced by `self.urls.extend(url)`

ernestask:
@mkutlak, `*urls` is not a tuple of URLs, it’s a tuple of positional arguments, so, while you might not ever get `([url, url], not url)`, you will always get `([url, url],)`, which is definitely not a tuple of URLs. One need only try to run the code to see that it does not work. I did just that, so this (or something else) is necessary.

ernestask:
```
NFO:faf.RepoSync:Processing repo 'rhel7-compose' assigned with OpSysRelease
INFO:faf.RepoSync:Processing repository 'rhel7-compose-7.6-x86_64' URL: '([u'http://abrt.brq.redhat.com/faf-repos/rhel7-compose-x86_64/'],)'
CRITICAL:faf:Action failed unexpectedly: AttributeError: 'list' object has no attribute 'endswith'
Traceback (most recent call last):
  File "/usr/bin/faf", line 137, in <module>
    main()
  File "/usr/bin/faf", line 117, in main
    exitcode = cmdline.func(cmdline, db)
  File "/usr/lib/python2.7/site-packages/pyfaf/actions/reposync.py", line 109, in run
    repo_instance['instance'].list_packages(list(architectures.keys()))
  File "/usr/lib/python2.7/site-packages/pyfaf/repos/rpm_metadata.py", line 263, in list_packages
    primaryfilename = self._get_primary_file_path(str(c), u)
  File "/usr/lib/python2.7/site-packages/pyfaf/repos/rpm_metadata.py", line 208, in _get_primary_file_path
    "repodata/repomd.xml")
  File "/usr/lib/python2.7/site-packages/pyfaf/repos/rpm_metadata.py", line 159, in _get_repo_file_path
    url = os.path.join(repourl, remote)
  File "/usr/lib64/python2.7/posixpath.py", line 77, in join
    elif path == '' or path.endswith('/'):
AttributeError: 'list' object has no attribute 'endswith'
```

mkutlak:
Oh, I see. We pass a list of urls here:
https://github.com/abrt/faf/blob/f796e816b2e93bb9af75d4f454d6355505eef7c3/src/pyfaf/actions/reposync.py#L90
(that's why you get the `([url,url], not url)`)

Now it makes sense. 

You could also 'fix' it with 
```python3
*[url.url for url in repo.url_list]
```
(which will unpack the list - [Unpacking Argument Lists](https://docs.python.org/3/tutorial/controlflow.html#tut-unpacking-arguments))

ernestask:
Took a different approach. Made passing URLs to `RpmMetadata` mandatory and kept the direct assignment, which called for some tweaking in a test.

ernestask:
Couldn’t find any instantiations without passing at least an empty list, so this should be fine in that regard.

mkutlak:
Good job :+1:

PR #784, opened 2019-04-24
Small fix to supress an error on user creation in the DB container

michalfabik:
`make run_db` in the `docker` directory was complaining about an existing user when run more than once.


xsuchy:
Can you instead call `createuser(1)`?

PR #792, opened 2019-05-28
Update Dockerfile for db

mkutlak:
- Change base image to registry.fedoraproject.org/f29/postgresql
- Update the base image during building

### DIFF
--- a/docker/Dockerfile_db
+++ b/docker/Dockerfile_db
@@ -1,4 +1,4 @@
-FROM candidate-registry.fedoraproject.org/postgresql
+FROM registry.fedoraproject.org/f29/postgresql
 
 USER root
 


mkutlak:
Updating (`dnf update`) the base image f**k ups the permissions.

ernestask:
> Updating (`dnf update`) the base image f**k ups the permissions.

Merged.

docker/Dockerfile_db
```diff
@@ -1,4 +1,4 @@
-FROM candidate-registry.fedoraproject.org/postgresql
+FROM registry.fedoraproject.org/f29/postgresql
```
ernestask: I wonder why the image isn’t updated anymore. FWIW, it’s also in the candidate registry: https://candidate-registry.fedoraproject.org/v2/_catalog, but they’re identical, so shouldn’t matter if we’re targeting a specific release.

PR #795, opened 2019-06-11
Keep supported CentOS versions in plugins/centos.conf

michalfabik:
See commit.

### DIFF
--- a/config/plugins/centos.conf
+++ b/config/plugins/centos.conf
@@ -1,3 +1,5 @@
 [CentOS]
 base-repo-url=http://vault.centos.org/centos/$releasever/os/Source/
 updates-repo-url=http://vault.centos.org/centos/$releasever/updates/Source/
+inactive-releases=
+active-releases=7
--- a/src/pyfaf/opsys/centos.py
+++ b/src/pyfaf/opsys/centos.py
@@ -17,6 +17,8 @@
 # along with faf.  If not, see <http://www.gnu.org/licenses/>.
 from __future__ import absolute_import
 
+import re
+
 from pyfaf.opsys import System
 from pyfaf.checker import DictChecker, IntChecker, ListChecker, StringChecker
 from pyfaf.common import FafError, log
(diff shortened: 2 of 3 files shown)


src/pyfaf/opsys/centos.py
```diff
@@ -187,7 +193,20 @@ def save_ureport(self, db, db_report, ureport, packages, flush=False, count=1):
...
+
+        for release in re.split(r"\s*[,;]\s*|\s+", self.inactive_releases):
+            result[release] = {
+                "status": "EOL",
+                "shortname": "centos-{0}".format(release)
+            }
+        for release in re.split(r"\s*[,;]\s*|\s+", self.active_releases):
+            result[release] = {
+                "status": "ACTIVE",
+                "shortname": "centos-{0}".format(release)
```
xsuchy: why the shortname? it was not previously in that dictionary.

michalfabik: I suppose i just wanted to keep it consistent with the [way its done for Fedora](https://github.com/abrt/faf/blob/f18b566afa2abfbd61d9278a44b6488b36d10430/src/pyfaf/opsys/fedora.py#L229), but you're right that it doesn't seem to be needed for either of the systems at the moment.

ernestask: I think it’s a derelict from the PkgDB days, but even then it’s not quite clear what it was for.

michalfabik: I see. Should I remove it from the Fedora part as well while I'm at it?

ernestask: I can’t see any harm coming from doing it, so why not.

config/plugins/centos.conf
```diff
@@ -1,3 +1,5 @@
 [CentOS]
 base-repo-url=http://vault.centos.org/centos/$releasever/os/Source/
 updates-repo-url=http://vault.centos.org/centos/$releasever/updates/Source/
+inactive-releases=5, 6
```
mkutlak: With this the 5 and 6 are added to FAF (even if as EOL) and we start to process uReports from them. And we not really want that. _(If uReport comes from unknown OpSys, it is rejected and moved to `deferred` dir.)_


So I think the `incactive-releases` should be empty by default.

michalfabik: removed

michalfabik: removed

ernestask: removed

PR #796, opened 2019-06-24
Merge pull request #796 from mgrabovsky/user-mail-typo

mgrabovsky:
This fixes an error when trying to download user data via `/download_data` when `EVERYONE_IS_ADMIN` is set, possibly among other things.


mkutlak:
Just for a reference:
https://github.com/abrt/faf/blob/7843c9ed8639fd5aa3a11ac99bc06912e5fd079e/src/pyfaf/storage/user.py#L29
https://github.com/abrt/faf/blob/7681c78d3a4d1d6a4509bf49e4f7d93d11e49b3d/src/webfaf/webfaf_main.py#L171

PR #798, opened 2019-06-26
Add querying for unknown packages by semver and semrel

### LINKED ISSUE #632: Problem quering with semver does not work with unknown packages
When querying problems and either of `since/to` `version/release` is selected, only problems with known packages are found. This means, that none of the reports that are reported for unknown packages are found.

How to reproduce: 
1) Enable accepting of unknown packages
2) Report with package, that does not exist in DB (not in table `packages`)
3) Create-problems
4) Query problem and make sure that your new problem is there
5) Set since version to value, that the given package should fit into and notice that the problem is not present anymore

How to fix:
In `src/webfaf/problem.py` in the part of code where since_version is used add also quering from table `reportunknownpackages`.

michalfabik:
Add semver and semrel columns to `reportunknownpackages` table, tweak tests, migrate DB with Alembic, add query to `problems.py`

Resolves #632

### DIFF
--- a/src/pyfaf/opsys/centos.py
+++ b/src/pyfaf/opsys/centos.py
@@ -36,6 +36,7 @@ from pyfaf.storage import (Arch,
                            column_len)
 from pyfaf.repos import repo_types
 from pyfaf.utils.parse import str2bool
+from pyfaf.storage.custom_types import to_semver
 
 __all__ = ["CentOS"]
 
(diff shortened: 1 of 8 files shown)


xsuchy:
Please join index on `semver` and `semrel`. Otherwise it looks good.

xsuchy:
LGTM

PR #800, opened 2019-07-10
centos: Use findall to search for (in)active releases

mkutlak:
re.split produces an empty string ('') when the (in)active release variable is empty.
That causes creation of CentOS with empty ('') release.


michalfabik:
I think the `\d` is redundant since `\w` in itself means `[a-zA-Z0-9_]`.

mkutlak:
Deleted the `\d`.

PR #804, opened 2019-07-31
Celery issues

### LINKED ISSUE #724: Issues with Celery
__How to run celery tasks with faf container:__
1) Download `redis` docker image
```
docker pull redis:latest
```
2) Start `redis` container with `hostname` and opened port `6379`:
```
docker run --name faf-redis --hostname faf-redis -dit -p 6379:6379 redis
```
3) Create `faf-network` and connect `faf` and `faf-redis` containers to it:
```
docker network create faf-network
docker network connect faf-network faf
docker network connect faf-network faf-redis
```
4) Change `hostname` in celery configuration to point to the `faf-redis` container:
```
docker exec faf sed -ie 's/localhost/faf-redis/g' /etc/faf/plugins/celery_tasks.conf
```
5) Restart `faf` container 
```
docker restart faf
```

Lot of the actions is not working on the `faf/celery_tasks/` page.

I see a lot of `TypeError: can't pickle generator objects` in logs.

michalfabik:
Resolves #724. The immediate cause was the use of generator objects instead of lists for the `choices` argument in some action form elements and the use of CmdLineParser method `add_mutually_exclusive_group` which was used by two of the actions and which wasn't implemented in the ActionFormArgparser. In the end, the mutual exclusion was ensured directly in `run` method of the respective actions for the CLI and handled ex-post based on the action's failure in case of the web UI.  
  
Many of the form elements were reworked in the process, form validation was added where practical, validation conflict was solved between the run action form and the scheduler form, methods for adding the form elements (action arguments) were refactored.

### DIFF
--- a/src/pyfaf/actions/archadd.py
+++ b/src/pyfaf/actions/archadd.py
@@ -43,4 +43,5 @@ class ArchAdd(Action):
         return 0
 
     def tweak_cmdline_parser(self, parser):
-        parser.add_argument("NAME", nargs="+", help="name of new architecture")
+        parser.add_argument("NAME", nargs="+", validators=[("InputRequired", {})],
+                            help="name of new architecture")
--- a/src/pyfaf/actions/cleanup_packages.py
+++ b/src/pyfaf/actions/cleanup_packages.py
@@ -106,7 +106,7 @@ class CleanupPackages(Action):
 
     def tweak_cmdline_parser(self, parser):
         opsys_group = parser.add_argument_group("Required together")
-        opsys_group.add_argument("OPSYS", nargs='?', help="operating system")
-        opsys_group.add_argument("RELEASE", nargs='?', help="release")
+        opsys_group.add_opsys_pos_arg(helpstr="operating system")
+        opsys_group.add_opsys_release_pos_arg(helpstr="release")
         arch_group = parser.add_argument_group("Individual")
-        arch_group.add_argument("--arch", help="architecture")
+        arch_group.add_arch(helpstr="architecture")
--- a/src/pyfaf/actions/cleanup_packages.py
+++ b/src/pyfaf/actions/cleanup_packages.py
@@ -29,18 +29,8 @@ class CleanupPackages(Action):
     name = "cleanup-packages"
 
     def run(self, cmdline, db):
-        # in case we're using the web UI:
-        if not hasattr(cmdline, "dry_run"):
-            cmdline.dry_run = False
-
-        # nobody will write the full name
-        if cmdline.OPSYS == "rhel":
-            cmdline.OPSYS = "Red Hat Enterprise Linux"
-
-        # check if operating system is known
-        if not get_opsys_by_name(db, cmdline.OPSYS):
-            self.log_error("Selected operating system '{0}' is not supported."
-                           .format(cmdline.OPSYS))
+        if not cmdline.OPSYS and not cmdline.RELEASE and not cmdline.arch:
+            self.log_error("None of the arguments were specified.")
             return 1
 
         if (cmdline.OPSYS or cmdline.RELEASE) and cmdline.arch:
@@ -50,6 +40,10 @@ class CleanupPackages(Action):
         if cmdline.OPSYS and not cmdline.RELEASE:
             self.log_error("Missing RELEASE argument.")
             return 1
+
+        # in case we're using the web UI:
+        if not hasattr(cmdline, "dry_run"):
+            cmdline.dry_run = False
 
         if cmdline.OPSYS:
             # nobody will write the full name
@@ -112,12 +106,10 @@ class CleanupPackages(Action):
                         .filter(Package.build_id == build.build_id)
                         .all()):
                 self.delete_package(pkg, cmdline.dry_run)
-                self.delete_package(pkg)
         return 0
 
     def tweak_cmdline_parser(self, parser):
-        opsys_group = parser.add_argument_group("Required together")
-        opsys_group.add_opsys(positional=True, helpstr="operating system")
-        opsys_group.add_opsys_release(positional=True, helpstr="release")
-        arch_group = parser.add_argument_group("Individual")
-        arch_group.add_arch(helpstr="architecture")
+        #TODO: use two argument groups: OS and release, and architecture # pylint: disable=fixme
+        parser.add_opsys(positional=True, helpstr="operating system")
+        parser.add_opsys_release(positional=True, helpstr="release")
+        parser.add_arch(helpstr="architecture")
--- a/src/pyfaf/actions/cleanup_packages.py
+++ b/src/pyfaf/actions/cleanup_packages.py
@@ -106,7 +106,7 @@ class CleanupPackages(Action):
 
     def tweak_cmdline_parser(self, parser):
         opsys_group = parser.add_argument_group("Required together")
-        opsys_group.add_opsys_pos_arg(helpstr="operating system")
-        opsys_group.add_opsys_release_pos_arg(helpstr="release")
+        opsys_group.add_opsys(positional=True, helpstr="operating system")
+        opsys_group.add_opsys_release(positional=True, helpstr="release")
         arch_group = parser.add_argument_group("Individual")
         arch_group.add_arch(helpstr="architecture")
--- a/src/pyfaf/actions/cleanup_packages.py
+++ b/src/pyfaf/actions/cleanup_packages.py
@@ -29,8 +29,18 @@ class CleanupPackages(Action):
     name = "cleanup-packages"
 
     def run(self, cmdline, db):
-        if not cmdline.OPSYS and not cmdline.RELEASE and not cmdline.arch:
-            self.log_error("None of the arguments were specified.")
+        # in case we're using the web UI:
+        if not hasattr(cmdline, "dry_run"):
+            cmdline.dry_run = False
+
+        # nobody will write the full name
+        if cmdline.OPSYS == "rhel":
+            cmdline.OPSYS = "Red Hat Enterprise Linux"
+
+        # check if operating system is known
+        if not get_opsys_by_name(db, cmdline.OPSYS):
+            self.log_error("Selected operating system '{0}' is not supported."
+                           .format(cmdline.OPSYS))
             return 1
 
         if (cmdline.OPSYS or cmdline.RELEASE) and cmdline.arch:
@@ -102,6 +112,7 @@ class CleanupPackages(Action):
                         .filter(Package.build_id == build.build_id)
                         .all()):
                 self.delete_package(pkg, cmdline.dry_run)
+                self.delete_package(pkg)
         return 0
 
     def tweak_cmdline_parser(self, parser):
--- a/src/pyfaf/actions/opsysdel.py
+++ b/src/pyfaf/actions/opsysdel.py
@@ -25,28 +25,29 @@ class OpSysDel(Action):
 
 
     def run(self, cmdline, db):
-        opsys = get_opsys_by_name(db, cmdline.NAME)
+        for opsys in cmdline.OPSYS:
+            db_opsys = get_opsys_by_name(db, opsys)
 
-        if not opsys:
-            self.log_error("Operating system '{0}' not found"
-                           .format(cmdline.NAME))
-            return 1
+            if not db_opsys:
+                self.log_warn("Operating system '{0}' not found"
+                              .format(db_opsys))
+                continue
 
-        if opsys.releases:
-            self.log_error("Unable to delete operating system with associated"
-                           " releases. Following is the list of associated "
-                           " releases:")
-            for release in opsys.releases:
-                self.log_error(release)
+            if db_opsys.releases:
+                self.log_warn("Unable to delete operating system with associated"
+                              " releases. Following is the list of associated "
+                              " releases:")
+                for release in db_opsys.releases:
+                    self.log_warn(release)
 
-            return 1
+                continue
 
-        self.log_info("Removing operating system '{0}'".format(cmdline.NAME))
+            self.log_info("Removing operating system '{0}'".format(opsys))
 
-        db.session.delete(opsys)
-        db.session.flush()
+            db.session.delete(db_opsys)
+            db.session.flush()
 
         return 0
 
     def tweak_cmdline_parser(self, parser):
-        parser.add_argument('NAME', help='name of new operating system')
+        parser.add_opsys_pos_arg(multiple=True, helpstr="operating system to delete")
--- a/src/pyfaf/actions/opsysdel.py
+++ b/src/pyfaf/actions/opsysdel.py
@@ -50,4 +50,4 @@ class OpSysDel(Action):
         return 0
 
     def tweak_cmdline_parser(self, parser):
-        parser.add_opsys(positional=True, multiple=True, helpstr="operating system to delete")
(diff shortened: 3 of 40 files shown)


ernestask:
@trapas, retest this please.

ernestask:
test this please

src/webfaf/blueprints/celery_tasks.py
```diff
@@ -298,7 +336,7 @@ def action_run(action_name):
     pt_form = PeriodicTaskForm(request.form)
     if request.method == "POST":
         if request.values.get("run") and action_form.validate():
-            run_action.delay(action_name, action_form.to_cmdline_dict())
+            run_action(action_name, action_form.to_cmdline_dict())
```
mgrabovsky: What's the reason for running the task in-process instead of having a worker do it? This will likely lead to timeouts for long-running tasks.

src/webfaf/blueprints/celery_tasks.py
```diff
@@ -253,6 +253,20 @@ def add_repo_type(self, choices=None, helpstr=None): # pylint: disable=unused-ar
...
+    def add_ext_instance(self, multiple=False, helpstr=None): # pylint: disable=unused-argument
+        Field = SelectField
+        if multiple:
+            Field = SelectMultipleField
+        field = Field(
+            "External FAF instance",
+            choices=[(a[0], str(a[0]) + " " + a[1])
+                     for a in db.session.query(ExternalFafInstance).with_entities(
+                         ExternalFafInstance.id, ExternalFafInstance.name)
+                     .order_by(ExternalFafInstance.id)],
```
mgrabovsky: Could use some refactoring to make it a bit more readable, as well.

src/pyfaf/cmdline.py
```diff
@@ -191,6 +191,16 @@ def add_repo_type(self, choices=None, helpstr=None):
         """
         self.add_argument("--type", choices=choices, help=helpstr)
 
+    def add_ext_instance(self, multiple=False, helpstr=None):
+        """
+        Add the `INSTANCE_ID` positional argument for an external FAF instance.
+        """
+        nargs = None
+        if multiple:
+            nargs = "*"
```
mgrabovsky: What about `nargs = "*" if multiple else None` to make a bit shorter and Pythonic? (Also applicable elsewhere in this PR.)

src/pyfaf/actions/opsysdel.py
```diff
@@ -25,28 +25,29 @@ class OpSysDel(Action):
...
+            db_opsys = get_opsys_by_name(db, opsys)
 
-        if not opsys:
-            self.log_error("Operating system '{0}' not found"
-                           .format(cmdline.NAME))
-            return 1
+            if not db_opsys:
+                self.log_error("Operating system '{0}' not found"
+                               .format(db_opsys))
+                continue
```
mgrabovsky: Is it an error then if we proceed anyway? Would a warning above be more appropriate if the error is not fatal?

src/pyfaf/cmdline.py
```diff
@@ -166,6 +166,16 @@ def add_opsys_with_rel_pos_arg(self, multiple=False, required=False): # pylint:
 
         self.add_argument("OPSYS", nargs=nargs, help="operating system (with release)")
 
+    def add_opsys_rel_status(self, required=False): # pylint: disable=unused-argument
```
mgrabovsky: Which argument is not used?

src/pyfaf/actions/releaseadd.py
```diff
@@ -67,6 +67,5 @@ def run(self, cmdline, db):
 
     def tweak_cmdline_parser(self, parser):
         parser.add_opsys()
-        parser.add_opsys_release()
-        parser.add_argument("-s", "--status", default="ACTIVE",
-                            help="ACTIVE, UNDER_DEVELOPMENT or EOL")
+        parser.add_argument("--opsys-release", help="operating system release")
```
mgrabovsky: Why this instead of the call to `add_opsys_release()`?

tests/test_actions.py
```diff
@@ -76,11 +76,6 @@ def test_releaseadd(self):
         self.assertEqual(self.call_action("releaseadd", {
             "opsys": "FooBarOS",
         }), 1)
-        self.assertEqual(self.call_action("releaseadd", {
```
mgrabovsky: What is the reason for removing the tests?

src/pyfaf/actions/archadd.py
```diff
@@ -16,6 +16,8 @@
 # You should have received a copy of the GNU General Public License
 # along with faf.  If not, see <http://www.gnu.org/licenses/>.
 
+from wtforms import validators
```
mgrabovsky: I really don't like pulling wtforms inside the CLI. Is there not a way around that?

src/pyfaf/retrace.py
```diff
@@ -1,5 +1,5 @@
 import re
-import concurrent.futures as futures
+import concurrent.futures as futures #pylint: disable=useless-import-alias
```
mgrabovsky: Is pylint really complaining about this?

src/pyfaf/actions/cleanup_packages.py
```diff
@@ -112,12 +106,15 @@ def run(self, cmdline, db):
...
-        opsys_group = parser.add_argument_group("Required together")
-        opsys_group.add_opsys(positional=True, helpstr="operating system")
-        opsys_group.add_opsys_release(positional=True, helpstr="release")
-        arch_group = parser.add_argument_group("Individual")
-        arch_group.add_arch(helpstr="architecture")
+        #opsys_group = parser.add_argument_group("Required together")
+        #opsys_group.add_argument("OPSYS", nargs='?', help="operating system")
+        #opsys_group.add_argument("RELEASE", nargs='?', help="release")
+        #arch_group = parser.add_argument_group("Individual")
+        #arch_group.add_argument("--arch", help="architecture")
```
mgrabovsky: Remove the commented code, please.

src/webfaf/blueprints/celery_tasks.py
```diff
@@ -286,8 +403,11 @@ def action_run(action_name):
     pt_form = PeriodicTaskForm(request.form)
     if request.method == "POST":
         if request.values.get("run") and action_form.validate():
-            run_action.delay(action_name, action_form.to_cmdline_dict())
-            flash("Action {0} started successfully.".format(action_name), "success")
+            output = run_action(action_name, action_form.to_cmdline_dict())
```
mgrabovsky: What's the reason for running the task in-process instead of having a worker do it? This will likely lead to timeouts for long-running tasks.

src/webfaf/blueprints/celery_tasks.py
```diff
@@ -141,22 +143,74 @@ def add_argument(self, *args, **kwargs):
...
+        if positional:
+            arg_str = "OPSYS"
+
+            q = db.session.query(OpSys.name)
+            choice_lst = [(a[0], a[0]) for a in q]
+
+            if with_rel:
+                q = q.join(OpSysRelease)
+                q = q.with_entities(OpSys.name, OpSysRelease.version)
+                choice_lst = [(a[0] + " " + a[1], a[0] + " " + a[1]) for a in q]
```
mgrabovsky: Can you factor this out? It seems a bit too busy around here. It would seem better to construct the list (or generator) of strings above and construct only the pair here.

src/pyfaf/actions/releaseadd.py
```diff
@@ -66,7 +67,8 @@ def run(self, cmdline, db):
         return 0
 
     def tweak_cmdline_parser(self, parser):
-        parser.add_opsys()
-        parser.add_opsys_release()
-        parser.add_argument("-s", "--status", default="ACTIVE",
-                            help="ACTIVE, UNDER_DEVELOPMENT or EOL")
+        parser.add_opsys(required=True, helpstr="operating system")
+        parser.add_argument("--opsys-release", required=True,
```
mgrabovsky: Why this instead of the call to `add_opsys_release()`?

tests/test_actions.py
ernestask: I remember suggesting removing some of them when parsing the choices for some of the command-line options was moved to argparse, since that was causing exceptions to be thrown before the `call_action()` could return. At that point we would be testing argparse, so it made sense to not do that.

Not sure if that still holds, since that was a while ago.

src/pyfaf/actions/releaseadd.py
michalfabik: Because `add_opsys_release()` adds a SelectField with _existing_ releases to the web UI form. We're adding a whole new release here so we need a plain TextField to get its name/number/version.

src/pyfaf/retrace.py
michalfabik: It was (that's why I added the `disable` directive) though I can't remember when exactly. It could be either when running `make rpm` or when building the local docker image with `make build_local` in the `docker` directory.

src/pyfaf/actions/repomod.py
```diff
@@ -98,6 +98,6 @@ def tweak_cmdline_parser(self, parser):
 
         group = parser.add_mutually_exclusive_group()
         group.add_argument("--gpgcheck", action="store_true",
-                           help="enable gpg check for this repository")
+                           help="enable GPG check for this repository")
         group.add_argument("--nogpgcheck", action="store_true",
-                           help="disable gpg check for this repository")
+                           help="disable GPG check for this repository")
```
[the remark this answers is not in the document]
michalfabik: It's meant to modify an existing property: `--nogpgcheck` to disable GPG check for a repo that has it enabled and `--gpgcheck` for the opposite.  
  
If `--nogpgcheck` were the default, then disabling  GPG check for a repo that has it enabled would require running `faf repomod` with no arguments at all, i.e. don't modify anything and set GPG check to the default value. Which is neither very intuitive (running the program without arguments) nor consistent behaviour (should the thing be left untouched, or set to default?).

src/pyfaf/actions/releaseadd.py
mgrabovsky: Ah, ok, makes sense.

src/pyfaf/retrace.py
mgrabovsky: That's strange. I haven't been able to reproduce this, neither in faf nor on a reduced, standalone case.

src/pyfaf/cmdline.py
michalfabik: removed pylint directive

src/pyfaf/cmdline.py
michalfabik: rewritten

src/pyfaf/retrace.py
ernestask: I think `from concurrent import futures` could be used as well, no?

src/pyfaf/actions/repomod.py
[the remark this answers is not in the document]
mgrabovsky: Oh, I understand now. If it's actually an action modifying a property, it seems more prudent to have it in a form like `--gpg-check=yes/no` to enable/disable the check explicitly, thus rendering the juggling of mutually exclusive argument void. In the web UI, this could be a dropdown with options such as "Do not change", "Enable" and "Disable".

src/pyfaf/actions/archadd.py
michalfabik: Fixed a mildly ugly workaround (using `exec()`) to pass the validators as strings. The alternative would probably be to write a separate method for each validator and for each of the two parsers.

src/pyfaf/actions/opsysdel.py
michalfabik: Changed to warning.

src/pyfaf/actions/archadd.py
ernestask:
```
from wtforms import validators

for validator in kwargs.get('validators'):
    class = getattr(validators, validator)
    field_kwargs['validators'].append(class())
```
?

src/pyfaf/actions/repomod.py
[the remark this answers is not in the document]
michalfabik: Yeah, that could be done. @mkutlak, would reworking the `--(no)gpgcheck` arg to `faf repomod` like this break anything? Up until now, I was trying not to change the behaviour of the CLI.

src/pyfaf/actions/archadd.py
ernestask: Of course, you would still need to add WTForms validator class names from the CLI.

src/pyfaf/actions/repomod.py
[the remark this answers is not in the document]
mkutlak: @michalfabik No

src/pyfaf/actions/archadd.py
michalfabik: Thanks! How silly of me not to think of that.

src/pyfaf/actions/repomod.py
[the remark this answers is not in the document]
michalfabik: Done

src/pyfaf/retrace.py
michalfabik: Yeah, fixed.

src/pyfaf/actions/cleanup_packages.py
michalfabik: Removed. I left them in on purpose to remind anybody working on this in the future to implement the arg groups and their mutual exclusiveness on the wtforms level, but oh well.

ernestask: That really never works, same as TODO comments. Better create a new issue for it with detailed information.

src/webfaf/blueprints/celery_tasks.py
```diff
@@ -141,6 +141,12 @@ def add_argument(self, *args, **kwargs):
 
         self.F.argparse_fields[kwargs["dest"]] = kwargs
 
+    def add_argument_group(self, *args, **kwargs): # pylint: disable=unused-argument
+        return ActionFormArgGroup(self.F)
+
+    def add_mutually_exclusive_group(self, *args, **kwargs): # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
michalfabik: Again, I was counting on their use in the future. Removed for now.

src/webfaf/blueprints/celery_tasks.py
michalfabik: I made the DB lines into one and moved the DB queries out of the field constructor calls in the other methods that you commented on. I also tried construction the lists and the tuples in two steps like you suggested but it looked a little over the top to me.

src/webfaf/blueprints/celery_tasks.py
michalfabik: Honestly, I have no idea what to do with this. The only change I did here was to remove the `delay` method(?), which magically made the actions work. (Before that, clicking the 'Run once' button in the web UI didn't actually do anything.) I can't find any similar uses of `delay` in the code or any pointers as to what it does. @ernestask , @mkutlak, any ideas?

ernestask: https://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.delay

Seems that it runs the task asynchronously, IIUC.

michalfabik: Solved. The original code was actually correct, only it was relying on the faf-celery-worker service in the host, which wasn't running.

PR #809, opened 2019-08-14
pyfaf: Allow ignoring specific Fedora releases when pulling

mgrabovsky:


mgrabovsky:
retest this please

src/pyfaf/opsys/fedora.py
```diff
@@ -105,6 +106,7 @@ def __init__(self):
         self.pagure_url = None
         self.build_aging_days = None
         self.koji_url = None
+        self.ignored_releases = None
```
mkutlak:
``` java 
************* Module pyfaf.opsys.fedora                                                                                                                                                                                                        
pyfaf/opsys/fedora.py:369:23: E1133: Non-iterable value self.ignored_releases is used in an iterating context (not-an-iterable)
```

mgrabovsky: Meh, the variable should be a list once `load_config_to_self()` is finished below, but I don't blame pylint for missing it. Will change to default to `[]` here, as well.

mkutlak:
LGTM. :+1:

PR #810, opened 2019-08-19
actions: create_problems: Bypass ORM when unassigning reports

ernestask:
For more SPEED.

### DIFF
--- a/src/pyfaf/queries.py
+++ b/src/pyfaf/queries.py
@@ -1027,6 +1027,16 @@ def get_unassigned_reports(db, report_type, min_count=0):
         query = query.filter(st.Report.count >= min_count)
     return query.all()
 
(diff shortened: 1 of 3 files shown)


src/pyfaf/queries.py
```diff
@@ -1027,6 +1027,12 @@ def get_unassigned_reports(db, report_type, min_count=0):
         query = query.filter(st.Report.count >= min_count)
     return query.all()
 
+def unassign_reports(db, report_ids):
```
mkutlak: Can you add a description of this function, so I know what it does without looking at the code?

ernestask: Mmm… okay. Though it is quite obvious in conjunction with the method name.

ernestask: Added a docstring.

PR #812, opened 2019-09-02
docker: Do not fail local Docker build if pylint fails

mkutlak:
It takes a lot of time to build the Docker image. It is annoying when
the whole build fails at the end because of an issue found by pylint
and you have to rerun the build.

### DIFF
--- a/docker/Dockerfile_local
+++ b/docker/Dockerfile_local
@@ -23,7 +23,8 @@ USER faf
 ENV HOME /faf
 
 # Build faf
-RUN ./autogen.sh && \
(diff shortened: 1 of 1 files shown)


ernestask:
But you want to run the actual tests, right? Because, I’m thinking, we could massage the makefile into passing `--nocheck` to rpmbuild on demand.

mkutlak:
> But you want to run the actual tests, right? Because, I’m thinking, we could massage the makefile into passing `--nocheck` to rpmbuild on demand.

Yes, I want to know if there is something wrong but just not fail the build because of it.

ernestask:
Oh, right, I misread the diff.

Issue #806, opened 2019-08-01
Better logging


ernestask:
It’s actually really sad how little is logged. For any basic debugging, one can only modify the code in production, which obviously doesn’t work for ongoing tasks.

What also is missing, IMO, are timestamps. Without them, it’s rather difficult to tell if there are bottlenecks.

StykMartin:
I have same feeling about this. Most of the time when something needs to be debugged we dont have enough information

xsuchy:
@ernestask does @mkutlak's PR resolve this?

ernestask:
No, it only addresses the issue in some parts.

PR #813, opened 2019-09-02
Improvements to logging system

mkutlak:
I did this mainly because it was not possible to make any changes to format of logging messages without editing the code.
 
___
**Changes**
- Logging configuration from a file (`/etc/faf/faf-logging.conf`)
- Added DATETIME to logging messages:
```
[DATE TIME] LEVEL:NAME: MESSAGE
```
- Added THREADNAME for logging from threads:
```
[DATE TIME] LEVEL:NAME(THREADNAME): MESSAGE
```
- Lazy formatting of debugging logs
- Removed custom logger for Python 2.6
- Get indexes using enumerate in for loops
- Added some debugging messages 
- Compute len of arrays outside of for loops

The old format:
```
LEVEL:NAME: MESSAGE
```

Also see: https://docs.python.org/3/library/logging.html
Related to #806

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -4,7 +4,8 @@
 
 SUBDIRS = plugins templates
 
-config_DATA = faf.conf
+config_DATA = faf.conf \
+              faf-logging.conf
 
 if HAVE_SYSTEMD
   config_DATA += celery-beat-env.conf
@@ -29,5 +30,6 @@ logrotatedir = ${sysconfdir}/logrotate.d
 EXTRA_DIST = celery-beat-env.conf \
              celery-worker-env.conf \
              faf.conf.in \
+             faf-logging.conf \
              faf-web.conf.in \
              faf
--- a/src/pyfaf/actions/create_problems.py
+++ b/src/pyfaf/actions/create_problems.py
@@ -53,8 +53,7 @@ class CreateProblems(Action):
         empty_problems = get_empty_problems(db)
         self.log_info("Found {0} empty problems".format(len(empty_problems)))
         for db_problem in empty_problems:
-            self.log_debug("Removing empty problem #{0}"
-                           .format(db_problem.id))
+            self.log_debug("Removing empty problem #%d", db_problem.id)
             db.session.delete(db_problem)
         db.session.flush()
 
@@ -170,8 +169,7 @@ class CreateProblems(Action):
             if match > 0:
                 # Ratio of problems matched
                 match_metric = float(match)/len(db_reports)
-                self.log_debug("Found possible match #{0} ({1:.2f})"
-                               .format(db_problem.id, match_metric))
+                self.log_debug("Found possible match #%d (%.2f)", db_problem.id, match_metric)
                 matches.append((match_metric, db_reports, db_problem))
 
         return matches
@@ -199,8 +197,7 @@ class CreateProblems(Action):
         for problem in problems:
             i += 1
 
-            self.log_debug("[{0} / {1}] Processing cluster"
-                           .format(i, len(problems)))
+            self.log_debug("[%d / %d] Processing cluster", i, len(problems))
 
             reports_changed = True
             problem_id = reuse_problems.get(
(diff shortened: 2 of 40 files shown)


mkutlak:
Let me know if you have more suggestions what could be improved/changed for logging.

ernestask:
Just nitpicks right now: using tabs instead of manually banging out four or however many spaces.

src/pyfaf/config.py
```diff
@@ -87,6 +90,24 @@ def load_config():
...
+    """
+    Load main config file for logging.
+    """
+    main_log_config_file = None
+    fpath = os.path.join(MAIN_CONFIG_DIR, MAIN_LOG_CONFIG_FILE)
+
+    if CONFIG_LOG_FILE_ENV_VAR in os.environ:
+        fpath = os.environ[CONFIG_LOG_FILE_ENV_VAR]
+
+    if os.access(fpath, os.R_OK):
```
ernestask: One thing is that these checks are prone to TOCTOU bugs. Even https://docs.python.org/3.7/library/os.html#os.access provides a better alternative.

src/pyfaf/actions/create_problems.py
```diff
@@ -299,8 +294,7 @@ def _create_problems(self, db, problemplugin, #pylint: disable=too-many-statemen
             i = 0
             for db_report in db_reports:
                 i += 1
```
ernestask: In most cases like this, we could just do `for i, db_report in enumerate(db_reports, start=1)`.

ernestask:
Oh, and doing `len(whatever)` just for log messages in a loop can be slow, but doesn’t matter a whole lot from what I’ve observed.

ernestask:
I’ll add some links to places where we could add messages (which I did in prod, but those don’t survive too long) that are rather helpful with timestamps now that we have them.

ernestask:
For example:
https://github.com/abrt/faf/blob/ff904405c519b895e3ac58de7192cda5b8ec4d41/src/pyfaf/actions/create_problems.py#L265
https://github.com/abrt/faf/blob/ff904405c519b895e3ac58de7192cda5b8ec4d41/src/pyfaf/actions/create_problems.py#L266

https://github.com/abrt/faf/blob/ff904405c519b895e3ac58de7192cda5b8ec4d41/src/pyfaf/problemtypes/core.py#L226

I just added a bunch of “[problem plugin] Fooing bars” before any large-ish SQL query (or before handing something off to satyr). Printing the problem plugin is also not a bad idea, since we run a bunch of processes that write to the same file, so figuring out what’s what can be a bit of a game.

ernestask:
Otherwise I don’t have any more suggestions, this is great work!

config/Makefile.am
```diff
@@ -30,4 +31,5 @@ EXTRA_DIST = celery-beat-env.conf \
              celery-worker-env.conf \
              faf.conf.in \
              faf-web.conf.in \
+             faf-logging.conf \
```
mgrabovsky: Just a teeny tiny suggestion: Move the added line one up so that the list is alphabetically sorted.

ernestask: I approve this message.

mkutlak:
> Just nitpicks right now: using tabs instead of manually banging out four or however many spaces.

Done.

> Oh, and doing `len(whatever)` just for log messages in a loop can be slow, but doesn’t matter a whole lot from what I’ve observed.

Done.

> I’ll add some links to places where we could add messages (which I did in prod, but those don’t survive too long) that are rather helpful with timestamps now that we have them.

Done.

src/pyfaf/config.py
```diff
@@ -87,6 +89,29 @@ def load_config():
 
     return result
 
+def configure_logging():
+    """
+    Load and configure logging from a config file.
+    If the config file isn't available, fallback to basic logging.
```
ernestask: __***Extreme***__ nitpick: “fallback” is a noun, what you mean is “fall back”.

ernestask:
I’ll merge this a bit later.

PR #814, opened 2019-09-04
pyfaf: Don't crash when no action is specified

mgrabovsky:
Don't crash on a runtime error in case the user calls `faf` with no actions. Instead, output an error and print the help message.

### DIFF
--- a/src/bin/faf
+++ b/src/bin/faf
@@ -84,6 +84,11 @@ def main():
         print("ABRT Analytics v{}".format(pyfaf.__version__))
         sys.exit(0)
 
+    if not hasattr(cmdline, "func"):
+        log.error("No command specified")
+        cmdline_parser.print_help()
+        sys.exit(1)
+
     # Catching too general exception Exception
     # pylint: disable-msg=W0703
 


ernestask:
What about something like https://stackoverflow.com/questions/23349349/argparse-with-required-subparser?

ernestask:
With `action_parsers = self.add_subparsers(title="action", dest="action", required=True)`, I get
```
usage: ./bin/faf [-h] [-v] [--sql-verbose] [-d] [--dry-run] [-V] action
./bin/faf: error: the following arguments are required: action
```

ernestask:
With something like https://stackoverflow.com/a/4042861, we could have a more generic solution and print the help message for every error.

mgrabovsky:
Setting `action` to required works quite well. However, it doesn't play with the `--version` flag, which should just print the version and exit if supplied.

ernestask:
Ah, true.

mgrabovsky:
Also, just looking at it now, I'm not (yet) quite sure, how to print the help message of the specific subcommand on error (akin to dynamic dispatch). Though maybe I'm misunderstanding the command line parser.

ernestask:
https://docs.python.org/3/library/argparse.html#action

Could help with the version bit (the `version` action).

ernestask:
I also don’t understand the parser, so we’re on the same boat. Just throwing ideas around.

mgrabovsky:
Ooh, that's cool. I never noticed the `version` action before. I might get around to doing this properly soon.

PR #823, opened 2019-09-25
Streamline setup of FAF w/ Celery and Redis in containers

michalfabik:
This PR introduces several changes to reduce the number of steps necessary to set up FAF, with the web UI using Celery and Redis, with three docker containers - for FAF itself, the database with the repo, package and report data, and for Redis which acts as a message broker and backend for Celery. Instructions in HACKING.md are expanded accordingly.
  
### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -108,7 +108,33 @@ otherwise an official image can be run with
 
     $ make run
 
+Connection to the database container is already set up in the FAF container in the `PGHOST`, `PGUSER`, `PGPASSWORD`, `PGPORT` and `PGDATABASE` environment variables. Alternatively, it can be set up by editing `/etc/faf/faf.conf` in the FAF container. The database container's IP address can be found by running
+
+    $ docker inspect -f='{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' db
+
 FAF should be available at ```http://localhost:8080/faf/```
+
+You can log in to the FAF container by running
+
+    $ make sh
+
+#### Web UI with Celery
+
+Download redis docker image
+
+    $ docker pull redis:latest
+
+Start redis container with hostname and opened port 6379:
+
+    $ docker run --name faf-redis --hostname faf-redis -dit -p 6379:6379 redis
+
+Create faf-network and connect faf and faf-redis containers to it:
+
+    $ docker network create faf-network
+    $ docker network connect faf-network faf
+    $ docker network connect faf-network faf-redis
+
+The message broker and backend used by Celery (Redis, in our case) are already set up in the FAF container in the `RDSBROKER` and `RDSBACKEND` environment variables. Alternatively, they can be set up by editing `/etc/faf/plugins/celery_tasks.conf` in the FAF container, followed by container restart.
 
 ### Reporting into FAF
 1. Set a `URL` to your server in `/etc/libreport/plugins/ureport.conf`
--- a/docker/openshift/faf-template.yml
+++ b/docker/openshift/faf-template.yml
@@ -161,7 +161,7 @@ parameters:
   value: faf-database
 
 - displayName: 'Database Port'
-  description: 'Database port on which the database is running on.'
+  description: 'Port on which the database communicates.'
   name: PG_FAF_PORT
   required: true
   value: '5432'
@@ -171,3 +171,15 @@ parameters:
   name: FAF_VOLUME_CAPACITY
   required: true
   value: '1Gi'
+
+- displayName: 'Redis broker'
+  description: 'Address of the Redis message broker machine, with redis scheme, port number and db number'
+  name: RDS_BROKER
+  required: false
+  value: 'redis://faf-redis:6379/0'
+
+- displayName: 'Redis backend'
+  description: 'Address of the Redis backend, with redis scheme and port number and db number'
+  name: RDS_BACKEND
+  required: false
+  value: 'redis://faf-redis:6379/0'
--- a/docker/openshift/faf-template.yml
+++ b/docker/openshift/faf-template.yml
@@ -105,6 +105,10 @@ objects:
             value: ${PG_FAF_HOST}
           - name: PGPORT
             value: ${PG_FAF_PORT}
+          - name: RDSBROKER
+            value: ${RDS_BROKER}
+          - name: RDSBACKEND
+            value: ${RDS_BACKEND}
           imagePullPolicy: IfNotPresent
           ports:
           - containerPort: 8080
@@ -161,7 +165,7 @@ parameters:
   value: faf-database
 
 - displayName: 'Database Port'
-  description: 'Port on which the database communicates.'
+  description: 'Database port on which the database is running on.'
   name: PG_FAF_PORT
   required: true
   value: '5432'
--- a/src/pyfaf/celery_tasks/__init__.py
+++ b/src/pyfaf/celery_tasks/__init__.py
@@ -1,18 +1,20 @@
 import datetime
 import logging
+import os
 import munch
 
 from celery import Celery
 from celery.signals import task_postrun
 from pyfaf.actions import actions
 from pyfaf.config import config
+from pyfaf.common import get_env_or_config
 from pyfaf.storage import DatabaseFactory, TaskResult
 from pyfaf.utils.contextmanager import captured_output_combined
 
 
 celery_app = Celery("pyfaf_tasks",
-                    broker=config.get("celery_tasks.broker", ""),
-                    backend=config.get("celery_tasks.backend", ""))
+                    broker=get_env_or_config("celery_tasks.broker", "RDSBROKER", ""),
+                    backend=get_env_or_config("celery_tasks.backend", "RDSBACKEND", ""))
 
 db_factory = DatabaseFactory(autocommit=True)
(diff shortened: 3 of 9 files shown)


HACKING.md
```diff
@@ -108,8 +112,46 @@ otherwise an official image can be run with
 
     $ make run
 
+Log in to the FAF container
+
+    $ make sh
+
+Set up the database in `/etc/faf/faf.conf`, under `[Storage]`. Use the database container's IP address found above as `DBHost`, set `DBPort` to `5432`, `DBName` to `faf`, `DBUser` to `faf` and `DBPasswd` to `scrt`.
```
ernestask: Why not provide a snippet of the configuration as it would appear in the file? Referencing the IP address might be a bit awkward, but it would be less of a word salad.

HACKING.md
```diff
@@ -99,6 +99,10 @@ Such build image can be run the same way
 
     $ make run_db
 
+Find out the database container's IP address:
+
+    $ docker inspect db | grep -m 1 '\<IPAddress'
```
ernestask: https://docs.docker.com/engine/reference/commandline/inspect/ as an alternative.

ernestask:
Looks good overall.

HACKING.md
michalfabik: Rewritten.

HACKING.md
michalfabik: Fixed.

ernestask: You didn’t *have to* :D I believe that might result in more than one address being output in some cases as well, if I’m not mistaken (`-m1` in grep would help guard against that).

HACKING.md
```diff
@@ -108,8 +112,55 @@ otherwise an official image can be run with
...
+
+Change hostname in Celery configuration to point to the faf-redis container:
+
+    $ docker exec faf sed -ie 's/localhost/faf-redis/g' /etc/faf/plugins/celery_tasks.conf
+
+Restart faf container
+
+    $ docker restart faf
+
+Start the Celery worker service in the host to enable manual action execution from the UI
```
mkutlak: s/in/on/

I don't think this is what we want with containers.

HACKING.md
```diff
@@ -108,8 +112,55 @@ otherwise an official image can be run with
 
     $ make run
 
+Log in to the FAF container
+
+    $ make sh
+
+Set up the database in the `[Storage]` section of `/etc/faf/faf.conf` as follows:
```
mkutlak: Could this be part of `run_local` action?

michalfabik: Not sure if I understand. Do you mean to automate the config file manipulation by adding stuff to the `run_local` target in`docker/Makefile`?  
  
The config file `/etc/faf/faf.conf` has to be edited like this no matter if you've run `make run` or `make run_local` so I thought it made sense to put this section after both of these options are mentioned, rather than bundling it with one of them.

mkutlak: > Do you mean to automate the config file manipulation by adding stuff to the `run_local` target in`docker/Makefile`?

Yes.

> The config file `/etc/faf/faf.conf` has to be edited like this no matter if you've run `make run` or `make run_local`

Oh, then do it for both.

michalfabik: See now. I added one Makefile target to use along with what was `make run_local`. With `make run`, it doesn't work in the current form as the container is run as non-root, so you still have to do it by hand.

mkutlak: You don't have to edit the `/etc/faf/faf.conf` with the non-root container. The container is ran under user `faf` and there are `env` variables set to connect to the database (PGUSER, PGHOST, ...).

We could add this to Dockerfile_local and you wouldn't need to do the config changes.
```
echo 'Defaults    env_keep =  "PGHOST PGUSER PGPASSWORD PGPORT PGDATABASE"' >> /etc/sudoers.d/faf
```

michalfabik: Pardon my being thick but where exactly in `Dockerfile_local` should I add this? I couldn't get it to work.  
  
Also, is there a similar way to set up the redis container in `/etc/faf/plugins/celery_tasks.conf`? If not, is there a way to get root in the non-root FAF container (run with `make run`, rather than `make run_local`) so I can edit the file by hand?

mkutlak: > Pardon my being thick but where exactly in `Dockerfile_local` should I add this? I couldn't get it to work.

Below/Above the `sed` is probably the right place. (add `&& \` at the end)
https://github.com/abrt/faf/blob/92cfd721908756383965e429ef10676c2a02f350/docker/Dockerfile_local#L36
___
> Also, is there a similar way to set up the redis container in `/etc/faf/plugins/celery_tasks.conf`? If not, is there a way to get root in the non-root FAF container (run with `make run`, rather than `make run_local`) so I can edit the file by hand?

You will have to do the same trick as with the `postgres` and replace:
https://github.com/abrt/faf/blob/92cfd721908756383965e429ef10676c2a02f350/src/pyfaf/celery_tasks/__init__.py#L14
with:
https://github.com/abrt/faf/blob/92cfd721908756383965e429ef10676c2a02f350/src/pyfaf/common.py#L179
and then supply the env vars (`RDSBROKER`, `RDSBACKEND` or w/e) in:
https://github.com/abrt/faf/blob/92cfd721908756383965e429ef10676c2a02f350/docker/Makefile#L16
___
And maybe change the openshift template:
https://github.com/abrt/faf/blob/master/docker/openshift/faf-template.yml

HACKING.md
mkutlak:
```bash
#!/bin/sh

source /etc/faf/celery-beat-env.conf

case $1 in
start)
        /usr/bin/python3 -m celery multi start $CELERYD_NODES \
        -A $CELERY_APP --pidfile=${CELERYD_PID_FILE} \
        --logfile=${CELERYD_LOG_FILE} --loglevel="${CELERYD_LOG_LEVEL}" \
        $CELERYD_OPTS &
        ;;
esac
```

```bash
#!/bin/sh

source /etc/faf/celery-worker-env.conf

case $1 in
start)
        /usr/bin/python3 -m celery multi start $CELERYD_NODES \
        -A $CELERY_APP --pidfile=${CELERYD_PID_FILE} \
        --logfile=${CELERYD_LOG_FILE} --loglevel="${CELERYD_LOG_LEVEL}" \
        $CELERYD_OPTS &
        ;;
stop)
        /usr/bin/python3 -m celery multi stopwait $CELERYD_NODES \
        --pidfile=${CELERYD_PID_FILE} &
        ;;
reload)
        /usr/bin/python3 -m celery multi restart $CELERYD_NODES \
        -A ${CELERY_APP} --pidfile=${CELERYD_PID_FILE} \
        --logfile=${CELERYD_LOG_FILE} --loglevel="${CELERYD_LOG_LEVEL}" \
        $CELERYD_OPTS &
        ;;
esac
```

mkutlak: This could replace those `services`.

HACKING.md
michalfabik: Reworked, HACKING.md rewritten accordingly.

HACKING.md
michalfabik: Added the two "service" scripts in the container. The scheduler is still broken though, even though the `beat` service is running. Will look into it.

src/pyfaf/celery_tasks/__init__.py
```diff
@@ -1,18 +1,28 @@
...
 
 
+def _get_env_or_config(conf_s, env_s, default):
+    found = os.environ.get(env_s, None)
+    if not found:
+        found = config.get(conf_s, None)
+        if not found:
+            found = default
+    return found
+
```
mkutlak: This can be deleted I guess.

docker/openshift/faf-template.yml
```diff
@@ -171,3 +171,15 @@ parameters:
   name: FAF_VOLUME_CAPACITY
   required: true
   value: '1Gi'
+
+- displayName: 'Redis broker'
+  description: 'Address of the Redis message broker machine, with redis scheme, port number and db number'
+  name: RDS_BROKER
```
mkutlak: This should be added below line 107.
```
- name: RDSBROKER
  value: ${RDS_BROKER}
- name: RDSBACKEND
  value: ${RDS_BACKEND}
```

HACKING.md
```diff
@@ -108,8 +108,34 @@ otherwise an official image can be run with
 
     $ make run
 
+Connection to the database container is already set up in the FAF container in the `PGHOST`, PGUSER`, `PGPASSWORD`, PGPORT` and `PGDATABASE` environment variables. Alternatively, it can be set up by editing `/etc/faf/faf.conf` in the FAF container. The database container's IP address can be found by running
```
mkutlak: PGUSER and PGPORT are missing \`

docker/openshift/faf-template.yml
michalfabik: Yeah, I was just going to ask what else is missing in this file. Added.

src/pyfaf/celery_tasks/__init__.py
michalfabik: Deleted.

HACKING.md
michalfabik: Fixed.

PR #824, opened 2019-09-26
CentOS 8 repo URLs

mkutlak:


src/pyfaf/opsys/centos.py
```diff
@@ -206,8 +202,12 @@ def get_releases(self):
         return result
 
     def get_components(self, release):
-        urls = [repo.replace("$releasever", release)
-                for repo in [self.base_repo_url, self.updates_repo_url]]
+        if not self.repo_urls:
+            self.log_info("No repository URLs were found.")
+            return []
+
+        repo_urls = re.findall(r"[\S]+", self.repo_urls)
```
mgrabovsky: In 4d592651, I introduced a function for parsing lists so that could be used here. It may be a callback for `load_config_to_self()`, as well.

mkutlak: Applied. Thanks.

PR #827, opened 2019-10-10
pyfaf: Remove version_info checks

mkutlak:
No need to check for this. We removed Python2 support in the past.


ernestask:
Jawohl.

Issue #826, opened 2019-10-07
pylint issues on Fedora 31 and Rawhide


mkutlak:
## Fedora 31:
```
************* Module pyfaf.rpm
pyfaf/rpm.py:81:11: E1101: Module 'rpm' has no 'error' member (no-member)
************* Module pyfaf.problemtypes.kerneloops
pyfaf/problemtypes/kerneloops.py:209:21: E1101: Module 'satyr' has no 'Kerneloops' member (no-member)
pyfaf/problemtypes/kerneloops.py:224:20: E1101: Module 'satyr' has no 'KerneloopsFrame' member (no-member)
************* Module pyfaf.problemtypes.core
pyfaf/problemtypes/core.py:174:17: E1101: Module 'satyr' has no 'GdbThread' member (no-member)
pyfaf/problemtypes/core.py:178:20: E1101: Module 'satyr' has no 'GdbFrame' member (no-member)
pyfaf/problemtypes/core.py:196:25: E1101: Module 'satyr' has no 'GdbStacktrace' member (no-member)
pyfaf/problemtypes/core.py:686:31: E1101: Module 'satyr' has no 'GdbStacktrace' member (no-member)
************* Module pyfaf.problemtypes.python
pyfaf/problemtypes/python.py:134:21: E1101: Module 'satyr' has no 'PythonStacktrace' member (no-member)
pyfaf/problemtypes/python.py:139:20: E1101: Module 'satyr' has no 'PythonFrame' member (no-member)
************* Module pyfaf.problemtypes.ruby
pyfaf/problemtypes/ruby.py:132:21: E1101: Module 'satyr' has no 'RubyStacktrace' member (no-member)
pyfaf/problemtypes/ruby.py:137:20: E1101: Module 'satyr' has no 'RubyFrame' member (no-member)
************* Module pyfaf.problemtypes.java
pyfaf/problemtypes/java.py:138:17: E1101: Module 'satyr' has no 'JavaFrame' member (no-member)
pyfaf/problemtypes/java.py:158:17: E1101: Module 'satyr' has no 'JavaThread' member (no-member)
************* Module pyfaf.actions.create_problems
pyfaf/actions/create_problems.py:320:28: E1101: Module 'satyr' has no 'Distances' member (no-member)
pyfaf/actions/create_problems.py:323:35: E1101: Module 'satyr' has no 'Dendrogram' member (no-member)
************* Module pyfaf.repos.rpm_metadata
[...]

PR #833, opened 2019-10-24
Fix pylint issues for rawhide build

michalfabik:
Fix/disable issues from Pylint 2.4, see http://pylint.pycqa.org/en/latest/whatsnew/2.4.html

### DIFF
--- a/src/pyfaf/storage/jsontype.py
+++ b/src/pyfaf/storage/jsontype.py
@@ -18,5 +18,6 @@ class JSONType(sa.types.TypeDecorator):
     def process_literal_param(self, value, dialect):
         return value
 
+    @property
     def python_type(self):
         return json
--- a/pylintrc
+++ b/pylintrc
@@ -38,7 +38,7 @@ load-plugins=
 # W0201 = Attribute defined outside __init__
 # R0201 = Method could be a function
 # useless-object-inheritance - We need this if we want to support Python2
-disable=W0141,W0201,R0201,too-few-public-methods,invalid-name,missing-docstring,superfluous-parens,duplicate-code,useless-object-inheritance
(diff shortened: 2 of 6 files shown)


src/pyfaf/actions/shell.py
```diff
@@ -54,7 +54,7 @@ def any(obj):
             return session.query(obj).order_by(func.random()).first()
 
         try:
-            import IPython
+            import IPython # pylint: disable=import-outside-toplevel
```
ernestask: What if we just disabled it globally?

src/pyfaf/storage/jsontype.py
```diff
@@ -18,5 +18,5 @@ def process_result_value(self, value, dialect):
     def process_literal_param(self, value, dialect):
         return value
 
-    def python_type(self):
+    def python_type(self): # pylint: disable=invalid-overridden-method
```
ernestask: Probably should rather just include a decorator? https://github.com/sqlalchemy/sqlalchemy/blob/8906617cd180f45219de2dd34dff8284f168fa9a/lib/sqlalchemy/sql/type_api.py#L392

ernestask:
I would be for disabling warnings for non-toplevel imports in pylintrc, since sprinkling more disabler comments is stretching the definition of “fixing”.

src/pyfaf/actions/shell.py
michalfabik: Disabled.

src/pyfaf/storage/jsontype.py
michalfabik: Done.

michalfabik:
Resolves #826

PR #837, opened 2019-10-31
tests: rpm_metadata: Avoid races

ernestask:
Currently, the socket for the HTTP server is created when the thread
is started, requiring the following code to sleep for a bit. Even with
that, it seems that the test fails on COPR due to some requests being
refused. Not sure what exactly is going on, but smells like a timing
issue as well. This commit rewrites the test in a way that ensure that
the socket starts listening way before we start serving requests.

### DIFF
--- a/tests/test_rpm_metadata.py
+++ b/tests/test_rpm_metadata.py
@@ -8,14 +8,15 @@ import glob
 import time
 
 import sys
-import queue
 
+from collections import Counter
+from functools import partial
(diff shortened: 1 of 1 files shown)


ernestask:
Either it’s a false positive or this really does fix the test on COPR, because the last two builds succeeded (well, one failed due to pylint being stupid, but that’s another story).

mkutlak:
https://copr.fedorainfracloud.org/coprs/g/abrt/faf-el8-devel/build/1082560/

LGTM :+1:

PR #840, opened 2019-11-13
webfaf: templates: report: item: Fit title in one line

ernestask:
Currently, the formatted title ends up retaining all newlines and
leading whitespace, leading to something quite messy. This also makes
fetching the title from gnome-abrt a bit painful. With some
trimming, we can make it be a single line again.


ernestask:
For reference, this is what happens currently (just imagine that title fetching actually works):
Screenshot from 2019-11-14 09-14-16

I could probably replace the newlines, but then we’re still left with massive indents.

mgrabovsky:
Looks great.

PR #846, opened 2019-11-28
Add bash autocompletion for faf

mkutlak:
You won't have to remember all the commands from now on or type --help
to run faf commands.*

Unfortunately the autocompletion isn't lightning fast.

### DIFF
--- a/src/bin/Makefile.am
+++ b/src/bin/Makefile.am
@@ -1,3 +1,9 @@
 bin_SCRIPTS = faf faf-migrate-db
 
+bashcompdir = $(sysconfdir)/bash_completion.d
+
+dist_bashcomp_DATA = \
+    faf.bash_completion \
+    faf-migrate-db.bash_completion
+
 EXTRA_DIST = $(bin_SCRIPTS)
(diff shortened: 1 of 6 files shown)


ernestask:
Well, sure, since it has to spawn a Python interpreter. You can write them by hand, if you want. :p

Also won’t work on zsh (and possibly friends) without flipping some switch somewhere, but that’s a different problem.

src/bin/Makefile.am
```diff
@@ -1,3 +1,7 @@
 bin_SCRIPTS = faf faf-migrate-db
 
+bashcompdir = $(sysconfdir)/bash_completion.d
+dist_bashcomp_DATA = faf.bash_completion \
```
ernestask:
```suggestion
dist_bashcomp_DATA = \
    faf.bash_completion \
```

ernestask: Otherwise it’s a rather annoying amount of whitespace.

ernestask:
One of our highly-trained engineers will look into that.

Thank you for using ABRT, @mastyk, we strive to deliver the best solutions for Beaker.

mgrabovsky:
Looks fine to me.

PR #852, opened 2019-12-03
ureport_compat: Fix encoding during ureport1to2

mkutlak:
I've run into this problem where ureport1 kerneloops would crash the
save-reports action during conversion from ureport1 to ureport2.

`TypeError: argument 1 must be str, not bytes`


mkutlak:
ureport1.TXT
[a789ad53-808d-4a41-a655-d3cec53cc705.txt](https://github.com/abrt/faf/files/3916591/a789ad53-808d-4a41-a655-d3cec53cc705.txt)

mgrabovsky:
Ah, got that a few times, too.

PR #853, opened 2019-12-09
Only set up logging when __name__ == __main__

ernestask:
This avoids unpleasant surprises when some unrelated Python code
(say, retrace-server) sets up loggers of its own.

https://docs.python.org/3/library/logging.config.html#logging.config.fileConfig

disable_existing_loggers defaults to True and results in rather
self-explanatory behavior: any loggers set up prior to importing
pyfaf.common will no longer output anything.

### DIFF
--- a/src/bin/faf
+++ b/src/bin/faf
@@ -56,18 +56,18 @@ def fix_ld_library_path():
 
 fix_ld_library_path()
 
-# pylint: disable=wrong-import-position
-import pyfaf
-from pyfaf.cmdline import CmdlineParser
-from pyfaf.config import config
-from pyfaf.common import Plugin, log, load_plugins
-from pyfaf.storage import getDatabase
-from pyfaf.utils.parse import str2bool
-
 NAME_PARSER = re.compile(r"^faf-([^{0}]+)$".format(os.sep))
 
 
 def main():
+    # pylint: disable=wrong-import-position
+    import pyfaf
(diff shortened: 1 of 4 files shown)


ernestask:
This or `disable_existing_loggers=False` (https://docs.python.org/3/library/logging.config.html#logging.config.fileConfig).

With this, I reckon we should replicate the configurable logging magic in retrace-server. With `disable_existing_loggers=False`, existing loggers will use the FAF configuration.

ernestask:
(Not tested extensively)

michalfabik:
Seems to be working fine with retrace-server.

mkutlak:
Old:
`$ faf-migrate-db --downgrade 8ac9b3343649`
```
INFO:alembic.runtime.migration:Context impl PostgresqlImpl.
INFO:alembic.runtime.migration:Will assume transactional DDL.
INFO:alembic.runtime.migration:Running downgrade fd5dc71471cc -> 8ac9b3343649, set-pkg-name-to-256
```

New:
`$ faf-migrate-db --downgrade 8ac9b3343649`
```

```

ernestask:
Hmm… I’ll take a better look at how this interacts with logging in other modules.

mgrabovsky:
Has this been fixed? I don't seem to be getting any logs from faf-migrate-db at the latest commit either.

ernestask:
No, I just haven’t had the time or willpower to look at this again.

PR #854, opened 2019-12-10
Create-problems speed-up

mkutlak:
We can save some time if we use more workers (threads) to process the db_reports
concurrently.

The communication with the DB is slowing down the db_report processing.
So while some of the workers are loading reports from the DB, other workers can start
processing the loaded reports.

The number of workers is limited by 'pool_size' and 'max_overflow' settings of connection engine.
(Default is 5 for pool_size and 10 for max_overflow).

Default is 4 threads.

faf create-problems -w 8
faf create-problems --max-workers 8

### DIFF
--- a/src/pyfaf/actions/create_problems.py
+++ b/src/pyfaf/actions/create_problems.py
@@ -18,7 +18,10 @@
 
 from operator import itemgetter
 from collections import defaultdict
+from concurrent.futures import as_completed, ThreadPoolExecutor
+
 import satyr
+
 from pyfaf.actions import Action
 from pyfaf.problemtypes import problemtypes
 from pyfaf.queries import (get_problems,
@@ -46,7 +49,6 @@ class HashableSet(set):
 
 class CreateProblems(Action):
     name = "create-problems"
-
 
     def _remove_empty_problems(self, db):
         self.log_info("Removing empty problems")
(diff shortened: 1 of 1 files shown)


mkutlak:
There is a possible performance improvement with [multiprocessing](https://docs.python.org/3/library/concurrent.futures.html?highlight=concurrent#processpoolexecutor), but I've run into a problem with the [pickle](https://docs.python.org/3/library/pickle.html#module-pickle) and satyr's objects.

ernestask:
Yeah, I can imagine it being more racy. This looks good at a glance, too.

/me will whip out the magnifying glass later.

src/pyfaf/actions/create_problems.py
```diff
@@ -255,6 +269,7 @@ def _iter_problems(self, db, problems, db_problems, problems_dict,
 
     def _create_problems(self, db, problemplugin, #pylint: disable=too-many-statements
                          report_min_count=0, speedup=False):
+
```
ernestask: :eyes:

mkutlak: 🕶️

ernestask:
test this please

ernestask:
RetEST tHiS PLEasE

ernestask:
Deploying Santa’s helpers…

PR #859, opened 2020-01-09
Avoid zero unique report counts in summaries

### LINKED ISSUE #847: Unique reports 0
I've seen several Reports that contain 'Unique reports 0'.
I believe it's a bug and that the number should be always >= 1.

mgrabovsky:
When saving an incoming report, the number of unique reports in the corresponding daily/weekly/monthly summary was incremented if and only if the incoming report had `serial` of 1, i.e. it was the first report of its kind from the machine.

However, it may happen that the very first report gets lost, is delayed etc., in which case the report's `serial` field is greater than 1 and the uniqe count in the summary is set to 0 if the report is the first of its kind on the server.

This is now fixed by setting the unique count in a summary to 1 whenever we receive a new report in that period, ignoring its `serial`.

Lastly, add a database migration which sets the unique count to at least one for all the daily/weekly/monthly summaries which claim to have zero unique reports, but do have _some_ reports.

Fixes #847

### DIFF
--- a/src/pyfaf/storage/migrations/script.py.mako
+++ b/src/pyfaf/storage/migrations/script.py.mako
@@ -1,5 +1,5 @@
-# Copyright (C) 2019  ABRT Team
-# Copyright (C) 2019  Red Hat, Inc.
+# Copyright (C) 2020  ABRT Team
+# Copyright (C) 2020  Red Hat, Inc.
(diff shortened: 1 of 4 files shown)


ernestask:
Makes sense to me.

PR #865, opened 2020-01-23
Allow deletion of multiple repos with faf repodel

michalfabik:
Make `faf repodel` accept one or more `REPO` arguments, add `--all` option.

### DIFF
--- a/src/pyfaf/actions/repodel.py
+++ b/src/pyfaf/actions/repodel.py
@@ -16,6 +16,8 @@
 # You should have received a copy of the GNU General Public License
 # along with faf.  If not, see <http://www.gnu.org/licenses/>.
 
+import re
+
 from pyfaf.actions import Action
 from pyfaf.storage.opsys import Repo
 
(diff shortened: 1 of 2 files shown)


mgrabovsky:
Can we have some tests for this, too?

mkutlak:
IMO pattern would be much more useful.

On retrace.fp.org we have repos like:
```
fedora-30-x86_64-one
fedora-30-x86_64-two
fedora-30-x86_64-three
fedora-31-x86_64-one
fedora-31-x86_64-two
fedora-31-x86_64-three
centos-7-x86_64-one
centos-7-x86_64-two
centos-7-x86_64-three
```

It would be much better to have `repodel --pattern="fedora-30*"` than `repodel --all`.

michalfabik:
> IMO pattern would be much more useful.
> It would be much better to have `repodel --pattern="fedora-30*"` than `repodel --all`.  
  
Do we need the extra `--pattern` option? I rewrote it so the `REPO` argument(s) get treated as patterns outright.

mkutlak:
> Do we need the extra `--pattern` option? I rewrote it so the `REPO` argument(s) get treated as patterns outright.

No, this is better :+1:

michalfabik:
> Can we have some tests for this, too?

Some tests added. Not nearly as extensive as they could be, given the possibilities of wildcards.

src/pyfaf/actions/repodel.py
```diff
@@ -25,24 +27,40 @@ class RepoDel(Action):
...
-        self.log_info("Removing repository '{0}'".format(cmdline.REPO))
+                self.log_info("Removing repository '{0}'".format(repo))
 
-        db.session.delete(repo)
-        db.session.flush()
+                db.session.delete(repo)
+
+            db.session.flush()
+        else:
+            self.log_error("No matching repositories found")
```
mkutlak: Nitpick:
`log_info` (log_warn) instead of `log_error`?

michalfabik: Fixed.

mkutlak:
LGTM :+1:

PR #869, opened 2020-01-31
pyfaf: actions: Use os.scandir where appropriate

ernestask:
According to https://docs.python.org/3/library/os.html#os.scandir, it is
faster for use cases where we need information other than just the name.

### DIFF
--- a/src/pyfaf/actions/save_reports.py
+++ b/src/pyfaf/actions/save_reports.py
@@ -228,10 +228,12 @@ class SaveReports(Action):
                 newest_older_ctime = stat.st_ctime
 
         report_filenames = []
-        for fname in os.listdir(self.dir_report_incoming):
-            stat = os.stat(os.path.join(self.dir_report_incoming, fname))
-            if fname[0] != "." and stat.st_mtime > newest_older_ctime and stat.st_mtime <= now:
-                report_filenames.append(fname)
+        with os.scandir(self.dir_report_incoming) as iterator:
+            for entry in iterator:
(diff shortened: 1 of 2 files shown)


ernestask:
The downside is that we don’t enumerate directories ahead of time and lose information such as file count.

ernestask:
Didn’t test everything extensively, but what I tested was faster, e.g. `_save_reports_speedup()` with a worst-case scenario of 100k files (20 iterations each):

listdir: 7.221339994925074 s
scandir: 5.170784448971972 s

src/pyfaf/actions/save_reports.py
```diff
@@ -306,36 +308,33 @@ def handle_term(_, __):
     def _save_attachments(self, db):
         self.log_info("Saving attachments")
 
-        attachment_filenames = os.listdir(self.dir_attach_incoming)
+        with os.scandir(self.dir_attach_incoming) as iterator:
+            for i, entry in enumerate(iterator, start=1):
+                self.log_info("[%d / ??] Processing file '%s'" % (i, entry.path))
```
mkutlak: I am not a fan of "??".

We don't recieve that many attachments (from logs its ~500 per week).

So IMO just drop the enumeration and replace the message with:
```
Processing attachment file "b14b14b14"
```

ernestask: Good, that makes two non-fans.

ernestask: Changed.

PR #870, opened 2020-02-03
pull-releases: Skip EOLed releases

mkutlak:
With configuration option SupportEOL you can select whether you want to
pull and in that way "support" EOLed releases.

With this option switched off EOLed releases won't be added to the database and
any incoming uReport from that release will be rejected as not supported.

### DIFF
--- a/src/pyfaf/actions/pull_releases.py
+++ b/src/pyfaf/actions/pull_releases.py
@@ -83,6 +83,11 @@ class PullReleases(Action):
 
                 continue
 
+            if not opsys_plugin.eol and remote_status == "EOL":
+                self.log_debug("Skipping unsupported EOL release '%s' (%s)" %
+                               (release, remote_status))
+                continue
+
             self.log_info("Adding release '{0}' ({1})"
                           .format(release, remote_status))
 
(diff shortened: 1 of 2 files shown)


src/pyfaf/actions/pull_releases.py
```diff
@@ -83,6 +83,9 @@ def run(self, cmdline, db):
 
                 continue
 
+            if not opsys_plugin.eol and remote_status == "EOL":
```
ernestask: Isn’t this dead code because of the `continue` above?

ernestask: Oh, hurr, indentation.

ernestask: What about logging something in this case?

ernestask:
LGTM overall.

src/pyfaf/actions/pull_releases.py
mkutlak: debug is enough?

ernestask: Yes, sir.

PR #874, opened 2020-02-14
Fix case of problem types

mgrabovsky:
Problem types should be lowercase as defined in Satyr.

### DIFF
--- a/src/pyfaf/storage/report.py
+++ b/src/pyfaf/storage/report.py
@@ -172,16 +172,6 @@ class ReportBacktrace(GenericTable):
 
         return crashthreads[0].frames
 
-    def normalized(self):
-        result = self.btp_thread()
-
-        if self.report.type == "kerneloops":
-            result.normalize_kerneloops()
-        elif self.report.type == "userspace":
-            result.normalize_userspace()
-
-        return result
-
     def as_named_tuples(self):
         '''
         Return list of named tuples containing name, path,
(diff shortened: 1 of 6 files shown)


src/pyfaf/storage/report.py
```diff
@@ -175,9 +175,9 @@ def frames(self):
     def normalized(self):
```
mkutlak: You can remove the function completely as it isnt used anywhere and `btp_thread()` was removed in commit b79e74351a85ccbabde565f2397d36b4629f0b00

src/pyfaf/utils/storage.py
```diff
@@ -19,37 +19,30 @@
...
     """
     Return formatted `reason` of the crash according to report type `rtype`
     """
 
-    if rtype == "USERSPACE":
-        res = RE_SIGNAL.search(reason)
-        if res:
-            return "{0} in {1}".format(res.group(), function_name)
-
+    if rtype == "userspace":
```
mkutlak: I think this should be `core`.

https://github.com/abrt/faf/blob/c7927635e89189dd70bf840498121f6f7c6ab2f4/src/pyfaf/ureport_compat.py#L35

mgrabovsky: Yup, thanks.

mkutlak:
LGTM :+1:

PR #876, opened 2020-02-21
Ameliorate LOB encoding craziness

mgrabovsky:
Disallow implicit encoding and decoding of Unicode strings in `GenericTable`'s `get_lob()` and `save_lob()`. This requires callers to handle the coding the themselves and only supply byte strings.

### DIFF
--- a/src/pyfaf/storage/generic_table.py
+++ b/src/pyfaf/storage/generic_table.py
@@ -49,38 +49,31 @@ class GenericTableBase(object):
 
     # lob for Large OBject
     # in DB: blob = Binary Large OBject, clob = Character Large OBject
-    def get_lob(self, name, binary=True):
+    def get_lob(self, name):
         lobpath = self.get_lob_path(name)
 
         if not os.path.isfile(lobpath):
             return None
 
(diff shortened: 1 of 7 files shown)


src/pyfaf/storage/generic_table.py
```diff
@@ -49,38 +49,31 @@ def has_lob(self, name):
 
     # lob for Large OBject
     # in DB: blob = Binary Large OBject, clob = Character Large OBject
-    def get_lob(self, name, binary=True):
+    def get_lob(self, name):
         lobpath = self.get_lob_path(name)
 
         if not os.path.isfile(lobpath):
```
ernestask: Unrelated: I think we should just try to `open()` the thing and handle errors from there.

mgrabovsky:
> Right, forgot to ask about those unused lobsters. Do you know where they came from?

They seem to have come from the distant, fertile lands known as 1e958ac4 and 39bc462b (or thereabouts), but their streams dried out in f6e578de.

I'll add the genealogy to the commit message.

ernestask:
+🦞

PR #878, opened 2020-02-26
WIP: Truncate crash function in GUI Reports table

michalfabik:
With very long function names in crash reports (up to 1024 chars in the DB), the Crash function column in the table at http://.../faf/reports/ makes the table unreasonably wide. This PR truncates the function name to a more reasonable value.  
  
Dtto in detailed tables for individual reports, Problem table and page title. (W3C [recommends](https://www.w3.org/Provider/Style/TITLE.html) maximum title length at 64 chars. Page title limited at _roughly_ twice that.)

### DIFF
--- a/src/webfaf/static/css/style.css
+++ b/src/webfaf/static/css/style.css
@@ -1511,3 +1511,26 @@ li.empty-line:last-child {
 .cell-count {
   text-align: right;
 }
+
+.truncate {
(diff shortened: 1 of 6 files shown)


ernestask:
Can you make the commit message a bit wordier? E.g. what the context is, why 120 in particular.

PR #879, opened 2020-02-26
webfaf: Bundle moment.js again

michalfabik:
This PR reverts commit [90fcc087](https://github.com/abrt/faf/commit/90fcc08707822fe926daa30e9bbaa480db8c1e20) and replaces moment.js with a newer version of moment-with-locales.js to avoid running into dependency issues due to orphaned packages.


ernestask:
Needs Provides and Bundled lines.

michalfabik:
> Needs Provides and Bundled lines.

We tried, we decided not to bother just for this one file.

PR #880, opened 2020-02-27
Prevent exception when calling decode on None

michalfabik:
This PR fixes an exception caused by calling `decode()` as a member of [`NoneType`](https://github.com/abrt/faf/blob/master/src/pyfaf/storage/generic_table.py#L52), introduced with a change of src/pyfaf/storage/report.py in #876.

### DIFF
--- a/src/pyfaf/storage/report.py
+++ b/src/pyfaf/storage/report.py
@@ -70,7 +70,10 @@ class Report(GenericTable):
 
     @property
     def oops(self):
-        return self.get_lob('oops').decode('utf-8')
+        result = self.get_lob('oops')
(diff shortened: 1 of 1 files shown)


src/pyfaf/storage/report.py
```diff
@@ -70,7 +70,10 @@ def bugs(self):
 
     @property
     def oops(self):
-        return self.get_lob('oops').decode('utf-8')
+        try:
+            return self.get_lob('oops').decode('utf-8')
```
mkutlak: Why not just check return value of `get_lob` and only `decode` it if it is not `None`?

PR #881, opened 2020-03-05
webfaf: Bundle typeahead.js again

michalfabik:
This PR reverts commit 4d00d1f1 and replaces typeahead.js with a newer version to avoid running into dependency issues due to orphaned packages. See also #879.


ernestask:
If we’re bundling minified versions, then https://github.com/twitter/typeahead.js/blob/master/dist/typeahead.bundle.min.js would be a better choice.

michalfabik:
> typeahead.bundle.min.js would be a better choice.  
  
Replaced.

mkutlak:
lgtm

PR #882, opened 2020-03-05
Truncate crash functions

ernestask:
See commits.

This mainly truncates the various strings displayed in the web UI. Some visuals incoming as well.

### DIFF
--- a/src/webfaf/static/js/custom.js
+++ b/src/webfaf/static/js/custom.js
@@ -172,6 +172,43 @@ $(document).ready(function() {
       $('#advanced-filters').removeClass('hide');
       $(this).addClass('hide');
     });
+
+    const observer = new ResizeObserver(entries => {
+      for (let entry of entries) {
+        const container = entry.target.querySelector('.crash-fn');
+        const expander = entry.target.querySelector('.expander');
+        const expanded = $(container).hasClass('expanded');
+        if (expanded && (container.scrollHeight == expander.scrollHeight)) {
+          container.classList.remove('expanded');
+
+          return;
+        }
+        const showExpander = !expanded && (container.scrollHeight > container.clientHeight);
+        const showCollapser = expanded;
+        if (showExpander) {
+          expander.innerHTML = 'show more';
+        }
+        else if (showCollapser) {
+          expander.innerHTML = 'show less';
+        }
+
+        expander.hidden = !(showExpander || showCollapser);
+      }
+    });
+
+    document.querySelectorAll(".crash-fn-container").forEach(element => {
+      expander = element.querySelector('.expander');
+      if (expander == null) {
+        return;
+      }
+
+      expander.addEventListener('click', (event) => {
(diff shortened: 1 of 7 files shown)


ernestask:
Can’t upload webums: https://cloud.gnome.org/s/EC68xxexDyZTB7c

ernestask:
The previous attempt with horizontal limiting and ellipsization is too fiddly, given that we’re doing this all in a table, where we have to apply a bunch of styles, resulting in the crash function column width being controlled largely by the header, so expanding the text would yield a long and narrow one.

src/webfaf/static/js/custom.js
```diff
@@ -172,6 +172,43 @@ $(document).ready(function() {
...
+        }
+        else if (showCollapser) {
+          expander.innerHTML = '[-]';
+        }
+
+        expander.hidden = !(showExpander || showCollapser);
+      }
+    });
+
+    document.querySelectorAll(".crash-fn-col").forEach(element => {
```
michalfabik: s/crash-fn-col/crash-fn-container/

src/webfaf/static/js/custom.js
```diff
@@ -172,6 +172,43 @@ $(document).ready(function() {
...
+        const expanded = $(container).hasClass('expanded');
+        if (expanded && (container.scrollHeight == expander.scrollHeight)) {
+          container.classList.remove('expanded');
+
+          return;
+        }
+        const showExpander = !expanded && (container.scrollHeight > container.clientHeight);
+        const showCollapser = expanded;
+        if (showExpander) {
+          expander.innerHTML = '[+]';
```
michalfabik: Suggest using [show more] and [show less] instead. The [+] and [-] remind me of forums with Reddit-like layout where it's used to expand/collapse a subtree of posts/messages and it looks a little odd centered at the bottom like this. But maybe it's just me.

src/webfaf/static/js/custom.js
ernestask: Have you tried it with that?

ernestask: Oh, I did drop the class altogether.

src/webfaf/static/js/custom.js
ernestask: Looking for analogies here is a bit pointless, but I can make it textual, sure. Should be fine, given that we don’t localize strings.

src/webfaf/templates/reports/item.html
```diff
@@ -10,7 +10,7 @@
 
 {% block title -%}
   {{ "Report #%d - %s" | format(report.id, component.name) -}}
-  {{ " in %s" | format(report.backtraces[0].crash_function) if report.backtraces | length > 0 }}
+  {{ " in %s" | format(report.backtraces[0].crash_function) | truncate(80, True) if report.backtraces | length > 0) }}
```
mkutlak: unexpected ')'

PR #890, opened 2020-03-17
Prepare project for use with tito

ernestask:
This commit makes building and releasing simpler by not requiring you to run arcane scripts that generate the spec file from a template just so you could get at the build dependencies. Complete with a Copr releaser.


ernestask:
Additionally, this would allow integrating with Packit easier.

ernestask:
The process can later be documented in the Wiki here to not have an overly crowded HACKING.md or adding yet another file.

.tito/custom/abrt/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+        files = [
+            metadata_file,
+            os.path.join(self.full_project_dir, self.spec_file_name),
+            'configure.ac',
+        ]
+        run_command('git add -- %s' % (' '.join([shlex.quote(file) for file in files])))
+
+        message = 'Release version %s' % (version)
+
+        run_command('git commit --message="%s"' % message)
```
ernestask: Fwiw, I’m hardcoding these things, because tito likes adding the release number to everything and it doesn’t make much sense to read the config in that case.

configure.ac
```diff
@@ -9,8 +9,10 @@ AC_DEFUN([FAF_PARSE_WITH],
     fi
     m4_popdef([FAF_UC_PACKAGE])])
 
+m4_define([faf_version], [2.2.0])
```
ernestask: Argh, forgot that this won’t get touched during builds, so the question is: do we care about being super-specific in development builds? I don’t think it matters at all for faf_schema, but maybe someone prefers seeing the commit hash in webfaf (and I don’t mean myself).

ernestask:
Wrote https://github.com/abrt/faf/wiki/Release-Checklist, but I’m going to automate Copr builds and add Packit for PR builds.

mkutlak:
And also update:
https://github.com/abrt/faf/blob/master/HACKING.md

ernestask:
Oh, yes, thanks.

mkutlak:
Building dockerfile locally is also broken now.

mkutlak:
``` bash
$ tito tag --use-version 2.2.1
Creating output directory: /tmp/tito
Tagging new version of faf: untagged -> 2.2.1-1
Traceback (most recent call last):
  File "/usr/bin/tito", line 23, in <module>
    CLI().main(sys.argv[1:])
  File "/usr/lib/python3.7/site-packages/tito/cli.py", line 203, in main
    return module.main(argv)
  File "/usr/lib/python3.7/site-packages/tito/cli.py", line 671, in main
    return tagger.run(self.options)
  File "/usr/lib/python3.7/site-packages/tito/tagger/main.py", line 114, in run
    self._tag_release()
  File "/home/mkutlak/ABRT/faf/.tito/custom/abrt/__init__.py", line 26, in _tag_release
    version = self._bump_version()
  File "/home/mkutlak/ABRT/faf/.tito/custom/abrt/__init__.py", line 16, in _bump_version
    with FileInput(files='configure.ac', inplace=True) as input:
NameError: name 'FileInput' is not defined
```

configure.ac
mgrabovsky: @ernestask Approximately how difficult would it be to make the git commit appear in the footer with the current setup? Asking for a friend.

ernestask: Tree-fiddy, approximately.

It can be baked in fairly easily, but the problem then is with tito builds, because it uses git-archive, meaning that the git metadata would not be there, requiring a custom builder. Unless you are into necromancy, which you could use to add additional mangling to the spec file. Still going to be fiddly, though.

mgrabovsky: Gotcha.

ernestask: So do you want me to look into it? If that’s genuinely helpul to you, I can figure something out.

ernestask: I mean your friend by saying “you”, of course.

mgrabovsky: No, not really, thanks. I was just wondering, because it was _fairly_ easy before, with autotools. At the moment, tito looks like an opaque monster to me, but I haven't started looking into it yet. I don't really need the information for anything in the UI.

PR #891, opened 2020-03-17
Set up GitHub Actions for running tests/pylint

ernestask:


ernestask:
Let’s see how well this works.

PR #896, opened 2020-03-22
Fix local Dockerfile build for tito changes

mkutlak:
#### Issue 1:
tito crashed with `_curses.error: setupterm: could not find terminal`
during the Dockerfile build.

Specifying TERM env variable works around the issue.

bugzilla:
bugzilla.redhat.com/show_bug.cgi?id=1733043

#### Issue 2:
- Install tito for build
- Use dnf builddep to install dependencies
- Change paths to rpms

### DIFF
--- a/container/Dockerfile_local
+++ b/container/Dockerfile_local
@@ -24,8 +24,7 @@ USER faf
 ENV HOME /faf
 
 # Build faf
-RUN sed -i '/^pylint/s/$/ ||:/' faf.spec && \
-    tito build --rpm --test
+RUN tito build --rpm --test
 
 #And continue as root
 USER 0
--- a/container/Dockerfile_local
+++ b/container/Dockerfile_local
@@ -13,9 +13,9 @@ WORKDIR '/faf'
 # Change owner of /faf, clean git and install dependences
 RUN chown -R --silent faf /faf && \
     chmod -R --silent g=u /faf && \
-    dnf -y install rpm-build sudo git which vim make &&  \
+    dnf -y install git-core make rpm-build sudo tito vim which && \
     git clean -dfx && \
-    eval dnf -y --setopt=strict=0 --setopt=tsflags=nodocs install $(./autogen.sh sysdeps) && \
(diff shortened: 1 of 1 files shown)


ernestask:
Fwiw, I’m trying something completely different, but rpmbuild is mysteriously failing.

ernestask:
But now it works after a kernel upgrade… Probably something with overlayfs.

container/Dockerfile_local
```diff
@@ -11,9 +13,9 @@ WORKDIR '/faf'
 # Change owner of /faf, clean git and install dependences
 RUN chown -R --silent faf /faf && \
     chmod -R --silent g=u /faf && \
-    dnf -y install rpm-build sudo git which vim make &&  \
+    dnf -y install git make rpm-build sudo tito vim which && \
```
ernestask:
```suggestion
    dnf -y install git-core make rpm-build sudo tito vim which && \
```

container/Dockerfile_local
```diff
@@ -22,15 +24,14 @@ USER faf
 ENV HOME /faf
 
 # Build faf
-RUN sed -i '/^pylint/s/$/ ||:/' faf.spec.in && \
-    ./autogen.sh && \
-    make rpm
+RUN sed -i '/^pylint/s/$/ ||:/' faf.spec && \
+    tito build --rpm --test
```
ernestask:
```suggestion
RUN tito build --rpm --test
```

container/Dockerfile_local
```diff
@@ -11,9 +13,9 @@ WORKDIR '/faf'
 # Change owner of /faf, clean git and install dependences
 RUN chown -R --silent faf /faf && \
     chmod -R --silent g=u /faf && \
-    dnf -y install rpm-build sudo git which vim make &&  \
+    dnf -y install git-core make rpm-build sudo tito vim which && \
```
mgrabovsky: Do you think it would be worth it to move this installation above `COPY . /faf` so that it leverage the cache better?

mkutlak: Ok, I did something. 

Please review.

PR #897, opened 2020-03-22
Rework container build

ernestask:
Currently, the “main” image is built by using RPMs from Copr, which
hurts reproducibility and automation, given that we have to build
everything in Copr first and only then build the container image. This
commit leaves only one Dockerfile file for production and development.
The image is built in multiple stages to prevent unwanted artifacts
staying around:

    1. Build FAF RPMs
    2. Upgrade system and install FAF RPMs
    3. Perform post-installation setup

Additionally, this reworks the Makefile shorthands: the names for
images, pods and volumes were changed to drop the redundant suffixes
(-image, -volume), the targets that depend on files existing now
actually do depend on them and common options were moved out to
variables.

### DIFF
--- a/container/Makefile
+++ b/container/Makefile
@@ -1,47 +1,65 @@
-POD_NAME = faf-pod
-VOLUME_NAME = faf-db-volume
+DB_IMAGE=abrt/faf-db
+LOCAL_IMAGE=abrt/faf-local
+IMAGE=abrt/faf
+
+DB_CONTAINER=faf-db
+CONTAINER=faf
+
+POD = faf-pod
+
+VOLUME = faf-db
+
 DB_PASS="NotSoSecretPassword"
+ENVIRONMENT = \
+	"-e PGHOST=localhost" \
+	"-e PGUSER=faf" \
+	"-e PGPASSWORD=scrt" \
+	"-e PGPORT=5432" \
+	"-e PGDATABASE=faf" \
+	"-e RDSBROKER=redis://faf-redis:6379/0" \
+	"-e RDSBACKEND=redis://faf-redis:6379/0"
 
-build:
-	podman build -t faf-image -f Dockerfile ../
-	podman tag faf-image abrt/faf-image
+build: Dockerfile
+	podman build --file=$< --tag=$(IMAGE) ../
 
-build_local:
-	podman build -t faf-image-local -f Dockerfile_local ../
+build_local: Dockerfile_local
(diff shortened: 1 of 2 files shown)


ernestask:
This is the last thing needed to completely automate releasing (container images built in Quay and Docker Hub, RPMs built in Copr).

ernestask:
Moreover, we can also think about building images for PRs, for easier testing.

container/Makefile
```diff
@@ -1,47 +1,65 @@
...
+	podman pod create -p 5432:5432 -p 6379:6379 -p 8080:8080 --name $(POD) || :
 
 run:
-	podman run --pod $(POD_NAME) --name faf -dit -e PGHOST=localhost -e PGUSER=faf -e PGPASSWORD=scrt -e PGPORT=5432 -e PGDATABASE=faf -e RDSBROKER=redis://faf-redis:6379/0 -e RDSBACKEND=redis://faf-redis:6379/0 abrt/faf-image
+	podman run --detach --interactive --pod $(POD) --name $(CONTAINER) --tty $(ENVIRONMENT) $(IMAGE)
 
-run_local:
-	podman run --pod $(POD_NAME) --name faf -dit -e PGHOST=localhost -e PGUSER=faf -e PGPASSWORD=scrt -e PGPORT=5432 -e PGDATABASE=faf -e RDSBROKER=redis://faf-redis:6379/0 -e RDSBACKEND=redis://faf-redis:6379/0 faf-image-local
+run_local: pod
+	podman run --detach --interactive --pod $(POD) --name $(CONTAINER) --tty $(ENVIRONMENT) $(LOCAL_IMAGE)
```
ernestask: Right, so, for development, this will be really slow. Not sure what would be the better strategy.

mkutlak:
I see one "issue"/disadvantage - you will be always `faf` inside the container.

So you won't be able to install extra packages for testing or do a quick config change.

ernestask:
> I see one "issue"/disadvantage - you will be always `faf` inside the container.
>
> So you won't be able to install extra packages for testing or do a quick config change.

If you really want, you can execute commands as an arbitrary user. In OpenShift, where you can’t do anything without rebuilding the image, I assume we should do the group ownership dance on config files as well. I imagine it’s currently not very easy to use anyway, because no one uses it and provides feedback.

ernestask:
Anyway, I’m just going to merge this. I only left in the changes to allow building the production image without needing to first build the packages in Copr (plus some tweaks to Makefile to make it more organized).

PR #899, opened 2020-03-25
webfaf: Refactor visualizations

mgrabovsky:
~Heavy work in progress for now.~

Improve the aesthetics of the plotting procedures as well as of the visuals themselves.

Ready for review now.


mgrabovsky:
Here's a sneak peek of the shiny new problem charts:

Screenshot_2020-03-31 ABRT Analytics - Problem #10762 - kernel in call_timer_fn

While at it, I also made some _a e s t h e t i c_ improvements. Here's a before and after on a sample report page:

onetwo

I have a to-do/wish list with a few items on it, so I'll keep digging a little in here.

mgrabovsky:
Ready for review now.

There's still some room for improvement, but I'll leave that for a later date.

mgrabovsky:
Before/after for the front page summary plot.

abrt-analytics-summary

src/webfaf/static/js/custom.js
```diff
@@ -64,6 +64,10 @@ $(document).ready(function() {
     });
 
     $('input.daterange').daterangepicker({
+      locale: {
+        // Set Monday as the first day of the week.
+        firstDay: 1
```
ernestask: Mmm… Maybe not. We should start using Moment.js with locales and use https://momentjs.com/docs/#/i18n/locale-data/.

ernestask: Actually, we would *only* need to use Moment.js with locales, as daterangepicker.js consults it for us.

mgrabovsky: I agree the UI should match the user's world view, but in this specific instance there's a slight twist – the weekly summaries as we store them assume Monday as the first day of the week. Letting the locale choose the first day might contradict this convention, so I'm not sure how beneficial it would be in in the end.

ernestask: So, with that in mind, the ability to choose the date range is just an exercise in futility?

https://github.com/abrt/faf/blob/5bc42f3c87ddae20ac01cb3d96e73f7a9c0d2770/src/webfaf/reports.py#L183

The above directly queries the database, so having the internal representation use ISO 8601 shouldn’t be an issue at all, given that `ReportHistoryWeekly` is already questionable in terms of usefulness (i.e. it only shows stats for calendar weeks, as opposed to letting me see trends in weekly intervals going back from *today*)<sup>*</sup>.

<sup>*</sup> - I wasn’t particularly fond of (or good at) statistics at school, so I cannot tell at all if what we show is useful or is just for the sake of having something.

src/webfaf/static/css/style.css
```diff
@@ -1447,7 +1447,7 @@ td > div.progress {
   margin-bottom: 0;
 }
 
-tr.odd, tr.stripe {
```
ernestask, line 1443: I was going to suggest going for `:nth-child(odd)`, but you would probably have to grab your can opener for the cans of spaghetti that would unleash.

ernestask:
One complaint: the legend in the summary page obscures part of the graph.

Screenshot from 2020-05-18 14-14-32

src/webfaf/static/js/custom.js
mgrabovsky: Well, the date range is exact when displaying daily values. It is aligned to Monday-starting weeks (as we store them in the DB) on weekly and months on monthly views.

I'm not sure, what you mean by "having the internal representation use ISO 8601 shouldn’t be an issue at all".

> it only shows stats for calendar weeks, as opposed to letting me see trends in weekly intervals going back from today

I understand your remark, but I _think_ it makes more sense that "weekly" mean calendar weeks rather than "7-day intervals going back from today".<sup>†</sup> Why not go forward by 7 days from the first day instead? The might be problems with months as well: Are they calendar months? Or 28/30/n-day intervals?

Calendar weeks starting with Mondays are technically unambiguous – of course, there are l10n differences among various regions.

And, naturally, in the end, the way we process, display and present data depends on the needs and preferences of whoever is _actually_ using ABRT Analytics for doing stuff and making decisions. Unfortunately, we currently don't have much feedback on this front.

† Putting aside the fact that there's a bug in the current UI where whatever dates the users selects in the date picker, they do not necessarily align with the actual weeks as they are stored/displayed when the page loads in a weekly view.

src/webfaf/static/css/style.css
mgrabovsky: Yup, it's unfortunately not that easy.

mgrabovsky:
> One complaint: the legend in the summary page obscures part of the graph.

Will fix that.

src/webfaf/static/js/custom.js
ernestask: Ah, right, I completely missed the dropdown for granularity (daily/weekly/monthly).

ernestask: Okay, whatever, let’s go with Mondays. I’m not willing to die on this hill.

mgrabovsky:
Should be OK now.

PR #903, opened 2020-03-27
Remove old hook for copr build

mkutlak:
Improvise. Adapt. Overcome.
Related: https://github.com/abrt/faf/pull/890

### DIFF
--- a/.copr/Makefile
+++ b/.copr/Makefile
@@ -1,7 +1,0 @@
-CDDIR=$(dir $(firstword $(MAKEFILE_LIST)))
-
-srpm:
-	dnf -y install autoconf automake autogen libtool
-	cd $(CDDIR)..;./autogen.sh; ./configure
-	make -C $(CDDIR).. srpm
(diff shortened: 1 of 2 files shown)


ernestask:
Oh. What is this even used for? I’ve changed the Copr projects to use tito.

mkutlak:
`@abrt/faf-devel` - we have a badge in REAME

ernestask:
So what? The build status is not affected by whatever is in those files.

ernestask:
In fact, I think this is what’s causing the failures.

ernestask:
So we should drop this and let Copr use Tito directly as configured in Copr itself.

mkutlak:
> So what? The build status is not affected by whatever is in those files.

What? The build status shows failing builds for

ernestask:
Wait… Oh ffs, we have multiple Copr projects.

mkutlak:
shoot, wrong button

Issue #909, opened 2020-05-12
faf attach-centos-bugs fails


djuarezg:
How to reproduce:

* Set up your account under `/etc/faf/plugins/centosmantisbt.conf`
* Run `faf create-problems`

This is the resulting error:

```
[2020-05-12 09:45:52] INFO:faf.AttachCentosBugs: Bug 15042 is not a valid ABRT issue.
[2020-05-12 09:45:52] INFO:faf.AttachCentosBugs: Processing Mantis issue #17140
[2020-05-12 09:45:52] DEBUG:zeep.transports: HTTP Post to https://bugs.centos.org/api/soap/mantisconnect.php:
<?xml version='1.0' encoding='utf-8'?>
<soap-env:Envelope xmlns:soap-env="http://schemas.xmlsoap.org/soap/envelope/"><soap-env:Body><ns0:mc_issue_get xmlns:ns0="http://futureware.biz/mantisconnect"><username>djuarezg</username><password>REDACTED</password><issue_id>17140</issue_id></ns0:mc_issue_get></soap-env:Body></soap-env:Envelope>
[2020-05-12 09:45:52] DEBUG:urllib3.connectionpool: Resetting dropped connection: bugs.centos.org
[2020-05-12 09:45:53] DEBUG:urllib3.connectionpool: https://bugs.centos.org:443 "POST /api/soap/mantisconnect.php HTTP/1.1" 200 5318
[2020-05-12 09:45:53] DEBUG:zeep.transports: HTTP Response from https://bugs.centos.org/api/soap/mantisconnect.php (status: 200):
<?xml version="1.0" encoding="UTF-8"?>
[...]

PR #912, opened 2020-05-13
webfaf: reports: Don’t call download_bug_to_storage_no_retry()

ernestask:
It is not part of the BugTracker interface and is not implemented by
Mantis. I’m not sure if taking the no-retry route was intentional,
however, so this commit assumes that it was a thinko.

Related to https://github.com/abrt/faf/issues/909.


ernestask:
Ah, yes, a possible explanation for `_no_retry` would be non-network-related exceptions causing a retry.

Issue #599, opened 2017-11-10
allow removing already associated bugs


vbenes:
when I accidentally associated NOTABUG and would like to remove it, no way.

I think we should see associated bugs and we should be able to remove them

thx

marusak:
This is currently WIP by an external student. I'll let you know about the progress.

anapauladsmendes:
Hello! I'm Ana Paula, I'm from Brazil and I'm a Outreachy applicant . I'm beginner, but I have been development my skills in Python, Selenium, Django, HTML, CSS, JavaScript, DOM and JQuery. I had some experience with Java in the University. I love learn and I am very excited to participate this apply with amazing people. I'm interested in this project and this issue. Could you explain more about the it?

marusak:
Hi @anapauladsmendes 
if you look at some problem, it may happen this problem has an external bug assigned. (for example https://retrace.fedoraproject.org/faf/problems/428460/)
It is possible to assign these problems if you are maintainer of the given package, but I don't think you can see this option on retrace.fedoraproject.org/faf but if you deploy your own FAF development container[1], by default you are admin there and therefore should be able to see button 'Associate bug' at any given report. If you assign such report you may realize you assigned a wrong one and therefore decide to remove it - however this is not implemented quite yet - and therefore this issue.

It would be nice to extend the page to which you are redirected after you click 'associate bug' button with list of already associated bugs and add some trash icon next to them to remove those. The specific details are left to the developer to propose solutions.

[1] https://github.com/abrt/faf/blob/master/HACKING.md#running-container-images

14Richa:
Hi @marusak any updates on this? I was looking to send my first PR. All other issues seems to have people working on them. @anapauladsmendes Have you made progress on this? Would you like us to work together maybe. Cheers :)

anapauladsmendes:
Hi, @14Richa I needed to abandon the Outreachy process now, you can solve this issue. Good luck!

laurenolivia:
@marusak Hi Matej, I'm an Outreachy applicant and recent graduate of a coding bootcamp. I see this issue may be taken and if so, could you please point me in the direction of a newcomer-friendly PR?

Prabhleen9Kaur:
@marusak Hi I am an outreachy aspirant i was looking for an issue to work as this issue is already taken up ,can you suggest me some other issue.

PR #913, opened 2020-05-13
webfaf: Disassociate bug from report

mkutlak:
**What's the problem?**
Right now there is no way how to dissociate a bug from report that was
added by mistake.
    
**How does this commit help?**
This commit adds a small button (trashcan) in front of links to
bugzillas. The button is only available (visible) to component maintainers.
They can use it to dissociate bugs from reports. Modal window appears after
clicking on the trashcan to confirm dissociation action.

<h1>Screenshots:</h1>

Button appears when report has an associated bug:
Button appears when report has an associated bug:

Modal after clicking on trashcan:
Modal after clicking on trashcan

Bug dissociated successfuly:
Bug dissociated successfuly

External bugs on Problem list page:
External bugs on Problem list page

Fixes: #599

### DIFF
--- a/src/webfaf/forms.py
+++ b/src/webfaf/forms.py
@@ -335,6 +335,10 @@ class AssociateBzForm(Form):
         (name, name) for name in bugtrackers])
 
 
+class DissociateBzForm(Form):
+    bug_id = BugIdField("Bug ID")
+
+
 class ProblemComponents(Form):
     component_names = StringField()
     submit = SubmitField("Reassign")
--- a/src/webfaf/reports.py
+++ b/src/webfaf/reports.py
@@ -70,7 +70,7 @@ from webfaf.utils import (Pagination,
                           is_component_maintainer)
 from webfaf.webfaf_main import db, flask_cache
 from webfaf.forms import (ReportFilterForm, NewReportForm, NewAttachmentForm,
-                          component_names_to_ids, AssociateBzForm)
+                          component_names_to_ids, AssociateBzForm, DissociateBzForm)
 
 reports = Blueprint("reports", __name__)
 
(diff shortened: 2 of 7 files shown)


src/webfaf/templates/reports/disassociate_bug.html
```diff
@@ -0,0 +1,28 @@
+{% extends "base.html" %}
+{% from '_helpers.html' import external_bugs %}
```
mgrabovsky, line 2: Inconsistent quote (single vs. double).

mkutlak: Fixed - changed to double quotes.

mkutlak:
> Looks good overall. Perhaps "dissociate" would be a better term, but the difference is most likely negligible.

Renamed because it's shorter. :)

mgrabovsky:
Hm, I'm thinking a much better UX would be to have the dissociation button/link next to the bug link under "External bugs". That would lead to a page with just a confirmation button (CSRF protection etc.). It would also remove the burden of having rewrite the bug number in the text field which should help avoid mistakes.

mkutlak:
> have the dissociation button/link next to the bug link under "External bugs". 

trashcan
So, button (trashcan) like this? (better styled)

mgrabovsky:
Yeah, that could work.

mgrabovsky:
@mkutlak Any progress here?

mkutlak:
I'll try to push new changes by end of the weekend.

mkutlak:
Broken tests:
```
extension "semver" has no installation script nor update path for version "0.22.0"
```

mkutlak:
The line with external bugs is ~2-3px taller than other lines. I've tried to fix it but I gave up, because it requires (IMO) bigger changes to FAF.

mgrabovsky:
Looks good, the icon is all right, but I'm just not keen on the button-y look (with the drop shadow) of the bug link. Do you think a simple link with more space to the right would work as well?

I just discovered a funny formatting error: 
image

mkutlak:
> I just discovered a funny formatting error:

Fixed.


> Do you think a simple link with more space to the right would work as well?

Like this?
change

mgrabovsky:
Yes, I think that looks better.

mkutlak:
### Problem page:
Screenshot_2020-06-27 ABRT Analytics - Problems

### Report page with delete button:
Screenshot_2020-06-27 ABRT Analytics - Report #3 - will-crash in __strlen_avx2

### Report page without delete button:
Screenshot_2020-06-27 ABRT Analytics - Report #3 - will-crash in __strlen_avx2(1)

Issue #919, opened 2020-05-19
Retrace cannot find packages even if downloaded


djuarezg:
**TL;DR: Retrace fails as it says that it cannot find source or binary packages (but it does find the debug ones).**

This one is a bit long but here are all steps on how to reproduce:

* Take a CentOS machine, configure `/etc/libreport/plugins/ureport.conf` with your server URL (`URL = https://abrtanalytics.web.cern.ch/faf` in our case).

* Add CentOS 7 release:
```
faf releaseadd -o centos --opsys-release 7 --status ACTIVE
```

* Configure repos that provide all involved packages:

```
sh-5.0$ faf repolist | awk '{print $1}' | xargs -L 1 faf repoinfo
Name: updates_debug
Type: dnf
URL: http://linuxsoft.cern.ch/cern/centos/7/updates/Debug/x86_64/
GPG check enabled: False
Name: updates_basic
Type: dnf
URL: http://linuxsoft.cern.ch/cern/centos/7/updates/x86_64/
GPG check enabled: False
Name: updates_source
Type: dnf
URL: http://linuxsoft.cern.ch/cern/centos/7/updates/Source/
GPG check enabled: False
Name: update_sources
Type: dnf
URL: http://linuxsoft.cern.ch/cern/centos/7/updates/Sources/
GPG check enabled: False
Name: os
Type: dnf
URL: http://linuxsoft.cern.ch/cern/centos/7/os/x86_64/
GPG check enabled: False
Name: os_source
Type: dnf
URL: http://linuxsoft.cern.ch/cern/centos/7/os/Source/
URL: http://linuxsoft.cern.ch/cern/centos/7/os/Sources/
GPG check enabled: False
```

* Assign repos to the previous release
```
faf repoassign <each_repo> "CentOS 7" x86_64
```
* Pull components

* Reposync only packages that will be relevant to the crash we will send:
```
faf reposync --name-prefix glibc -d -v
faf reposync --name-prefix coreutils -d -v
# This will start printing that debuginfo, source rpms and binary rpms are being downloaded, stored afterwards on /var/spool/faf/lob/Package/package/...
```

* Force a crash on the CentOS machine:

```
sleep 10m &
kill -SIGSEGV <ID>
```

* This will send the ureport to your `faf` server, run `faf save-reports` and `faf create-problems`.

* Run `faf retrace -v -d`:

```
[2020-05-19 09:38:22] INFO:faf.Retrace: Processing 'Crash of user-space binary' problem type
[...]

PR #920, opened 2020-05-22
repos: dnf: Correctly handle debug packages

ernestask:
5508c07b919a159e56acaad0bd192e02297152b8 completely ignored the problem
of base package names, resulting in debuginfo and debugsource packages
having distinct builds in the database and breaking retracing. While
there is no longer a property in DNF that spits out the base name, we
can use the class attributes and strip the suffixes as needed. This
commit does just that.

Fixes https://github.com/abrt/faf/issues/919


ernestask:
The code looks rather inefficient, so feel free to suggest something else.

PR #923, opened 2020-06-02
Annotate function return types

michalfabik:
See https://docs.python.org/3/library/typing.html  
  
This PR only adds type annotations. It doesn't deal with any existing type errors and incompatibilities e.g. as reported by `mypy`.


michalfabik:
Pushed a few fixes but https://github.com/abrt/faf/pull/923/commits/036c91d120b0c42518b87db1a98b0456f7be5909 is still broken.

src/pyfaf/faf_rpm.py
```diff
@@ -30,7 +31,7 @@
 __all__ = ["store_rpm_deps", "unpack_rpm_to_tmp"]
 
 
-def parse_evr(evr_string):
+def parse_evr(evr_string) -> Tuple[Optional[str], Optional[str], Optional[str]]:
```
ernestask, line 34: As you can see, epoch can also be set to a literal zero, making `Optional[str]` not completely true.

ernestask:
Somehow the tests run fine for me locally, not sure what the difference is.

ernestask:
> Somehow the tests run fine for me locally, not sure what the difference is.

Oh, I’m an idiot. I create a branch without your changes.

src/pyfaf/storage/problem.py
```diff
@@ -25,6 +27,7 @@
 from .generic_table import GenericTable
 from .opsys import Build, OpSysComponent, OpSysRelease
 from .user import User
+from .report import Report
```
ernestask, line 30: LGTM very clearly suggests that this might be the cause of failing imports.

src/pyfaf/storage/problem.py
```diff
@@ -133,44 +136,44 @@ def quality(self):
...
-    def archived_reports(self):
+    def archived_reports(self) -> List[Any]:
         '''
         List of archived reports
         '''
         return [r for r in self.reports if r.archived]
 
     @property
-    def backtraces(self):
+    def backtraces(self) -> List[Report]:
```
ernestask, line 161: How did you come to this conclusion?

src/pyfaf/storage/problem.py
```diff
@@ -133,44 +136,44 @@ def quality(self):
...
-    def active_reports(self):
+    def active_reports(self) -> List[Dict[str, Any]]:
         '''
         List of all non archived reports
         '''
         return [r for r in self.reports if not r.archived]
 
     @property
-    def archived_reports(self):
+    def archived_reports(self) -> List[Any]:
```
ernestask, line 154: How is this `List[Any]`? Isn’t `Problem.reports` a backref defined in `Report`, making this `List[Report]`?

src/pyfaf/storage/problem.py
ernestask: Shouldn’t this be `List[ReportBacktrace]`?

https://github.com/abrt/faf/blob/8b5af92f8915f15b8049822e397a5573ee901041/src/pyfaf/storage/report.py#L166

src/pyfaf/storage/problem.py
```diff
@@ -133,44 +136,44 @@ def quality(self):
...
-    def backtraces(self):
+    def backtraces(self) -> List[Report]:
         '''
         List of all backtraces assigned to this problem.
         '''
         return sum([x.backtraces for x in self.reports], [])
 
     @property
-    def sorted_backtraces(self):
+    def sorted_backtraces(self) -> List[Report]:
```
ernestask, line 168: Same thing here, you are sorting `Problem.backtraces`, which is `List[ReportBacktrace]`.

src/pyfaf/actions/create_problems.py
```diff
@@ -63,7 +65,7 @@ def _remove_empty_problems(self, db):
             db.session.delete(db_problem)
         db.session.flush()
 
-    def _get_func_thread_map(self, threads):
+    def _get_func_thread_map(self, threads) -> Dict[str, set]:
```
ernestask, line 68: Not `Set[ReportBtThread]`?

src/pyfaf/actions/create_problems.py
```diff
@@ -135,7 +137,7 @@ def _get_thread_map(self, func_thread_map, max_cluster_size):
 
         return thread_map
 
-    def _create_clusters(self, threads, max_cluster_size):
+    def _create_clusters(self, threads, max_cluster_size) -> List[List[Any]]:
```
ernestask, line 140: Why `Any` if you had annotated `_get_thread_map()`?

ernestask:
Plenty of room for improvement still. Be careful about what you import, too, because the code is organized as well as my life, so it can come crashing down *very* easily.

Issue #927, opened 2020-06-10
reposync fails parsing certain rpm dependencies


djuarezg:
I think this is a corner case we might be facing ourselves which might be too specific but I think package version parsing could be causing the following:

```
[2020-06-10 14:37:58] DEBUG:faf.pyfaf.faf_rpm: php-oracle-1.0-4.el7.cern.x86_64 contains 5 files
[2020-06-10 14:37:58] CRITICAL:faf: Action failed unexpectedly: DataError: (psycopg2.errors.InvalidTextRepresentation) invalid input syntax for integer: "%{epoch}"
LINE 1: ...25, 'REQUIRES', 'oracle-instantclient-basic', 12, '%{epoch}'...
                                                             ^

[SQL: INSERT INTO packagedependencies (package_id, type, name, flags, epoch, version, release) VALUES (%(package_id)s, %(type)s, %(name)s, %(flags)s, %(epoch)s, %(version)s, %(release)s) RETURNING packagedependencies.id]
[parameters: {'package_id': 2825, 'type': 'REQUIRES', 'name': 'oracle-instantclient-basic', 'flags': 12, 'epoch': '%{epoch}', 'version': '12.1', 'release': None}]
(Background on this error at: http://sqlalche.me/e/9h9h)
Traceback (most recent call last):
  File "/usr/lib64/python3.8/site-packages/sqlalchemy/engine/base.py", line 1283, in _execute_context
    self.dialect.do_execute(
  File "/usr/lib64/python3.8/site-packages/sqlalchemy/engine/default.py", line 590, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.InvalidTextRepresentation: invalid input syntax for integer: "%{epoch}"
LINE 1: ...25, 'REQUIRES', 'oracle-instantclient-basic', 12, '%{epoch}'...
                                                             ^


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/bin/faf", line 152, in <module>
    main()
  File "/usr/bin/faf", line 132, in main
    exitcode = cmdline.func(cmdline, db)
  File "/usr/lib/python3.8/site-packages/pyfaf/actions/reposync.py", line 240, in run
    res = store_rpm_deps(db, package, repo_instance['nogpgcheck'])
  File "/usr/lib/python3.8/site-packages/pyfaf/faf_rpm.py", line 133, in store_rpm_deps
    db.session.flush()
[...]

PR #930, opened 2020-06-15
pyfaf: faf_rpm: Don’t parse EVRs with non-numeric epochs

ernestask:
Closes https://github.com/abrt/faf/issues/927


ernestask:
Will rewrite the commit message for added clarity and actually fix the issue. I forgot to add some `continue` statements, but now I realize that it’s probably better to skip the package entirely and not the broken dependencies.

Issue #928, opened 2020-06-12
reposync does not take dynamic repos into account


djuarezg:
`faf reposync` does not take into account that repos can change their packges, i.e. a previously existing rpm could be removed from it.

There should be a mechanism to check repo metadata to actually refresh package lists, otherwise the following scenario could happen:

* Repo A has 10 packages (added as `dnf` type)
* Add repo A to FAF
* Repo A has now only 5 packages and its metadata has been refreshed
* FAF tries to reposync no longer existing packages from this repo, resulting in an error 404:

```
[2020-06-12 08:54:37] DEBUG:faf.RepoSync: [7 / 15] Processing package firefox-debuginfo
[2020-06-12 08:54:37] INFO:faf.RepoSync: Adding package firefox-debuginfo-68.9.0-1.el7.centos.x86_64.rpm
[2020-06-12 08:54:37] INFO:faf.RepoSync: Downloading http://linuxsoft.cern.ch/cern/centos/7/updates-testing/Debug/x86_64/firefox-debuginfo-68.9.0-1.el7.centos.x86_64.rpm

Exception occurred, retrying in 5 seconds 1/3
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/pyfaf/utils/decorators.py", line 47, in f_retry
    return func(*args, **kwargs)
  File "/usr/lib/python3.8/site-packages/pyfaf/actions/reposync.py", line 260, in _download
    pipe = urllib.request.urlopen(url)
  File "/usr/lib64/python3.8/urllib/request.py", line 222, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib64/python3.8/urllib/request.py", line 531, in open
    response = meth(req, response)
  File "/usr/lib64/python3.8/urllib/request.py", line 640, in http_response
    response = self.parent.error(
  File "/usr/lib64/python3.8/urllib/request.py", line 569, in error
    return self._call_chain(*args)
  File "/usr/lib64/python3.8/urllib/request.py", line 502, in _call_chain
    result = func(*args)
  File "/usr/lib64/python3.8/urllib/request.py", line 649, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 404: Not Found
```
* It can also happen the other way around, a recently added rpm may not be downloaded as it has been recently added to the repo.

[...]

PR #931, opened 2020-06-16
reposync: Allow expiring cache earlier

ernestask:
See commits.

Closes https://github.com/abrt/faf/issues/928


ernestask:
Needs tests and maybe some documentation somewhere.

ernestask:
Adding a test revealed some more DNF-related surprises, so this might take a bit more.

mgrabovsky:
Looks good to me at a glance, haven't tested locally yet.

Issue #918, opened 2020-05-18
Openshift deployment without requiring admin privileges


djuarezg:
Openshift deployment templates are really great, but could be adapted to run under an unprivileged environment, i.e. shared Openshift cluster where you might not have admin privileges for running Security context changes.

Currently the required workaround is to add `initContainers` + volumes or build images ourselves with permission edits.

djuarezg:
Related to this privileges templates could be improved by separating configuration files to configMaps, as in this extract:

```
volumeMounts:
          (...)
          - mountPath: /etc/faf/plugins/centosmantisbt.conf
            name: centosmantis
            subPath: centosmantisbt.conf
        volumes:
        (...)
        - configMap:
            name: centosmantis
          name: centosmantis
```

mkutlak:
IMO the solution in https://github.com/abrt/faf/commit/f06548b4a72f60a8b2e0788d9390916ade9b4dd8 is not for OpenShift/Kubernetes. 

If a pod with faf would be killed, you would lose all your configurations because the `/etc/faf` directory is not stored on persistent volume.

So this issue should be reopened.

PR #940, opened 2020-09-03
container: Make configuration files writable by root group

ernestask:
With rootless containers, the configuration is essentially immutable.
That can be worked around by making it group-writable (the faf user will
have membership of the root group).

Closes https://github.com/abrt/faf/issues/918


mgrabovsky (approved):
Makes sense to me.

Issue #922, opened 2020-05-27
Attaching MantisBT bugs to reports does not work, hashes differ depending on FAF server


djuarezg:
This issue is derived from https://github.com/abrt/faf/issues/523#issuecomment-634046143

`faf attach-centos-bugs` links reports with already existing MantisBT bugs  based on the hash that appears on the `abrt_hash` field. The problem is that the `abrt_hash` differs depending to to which server the ureport is being sent.

Same error, reported to two different instances, creates a different `bthash`:

- https://retrace.fedoraproject.org/faf/reports/bthash/058896b486386a0060459830992f69a275e7f3fd
- https://abrtanalytics.web.cern.ch/faf/reports/bthash/5bb483b5ca0ea992b911f4d845479a085eab0f24

```
faf=# select * from reportbthashes;
 type  |                   hash                   | backtrace_id 
-------+------------------------------------------+--------------
 NAMES | 0a71a4adf9b2efcb9d6608d6568c1e6836f15574 |            1
 NAMES | d2c00324955eeb5973b0f37b53aab7086faf7144 |            1
(2 rows)
```

```
[djuarezg@localhost ~]$ abrt-cli list
Private Reports is enabled, use 'abrt-cli -a COMMAND' to get the detected problems.
id 6e108d1053e4e12b96d7db1dab05205fa6318293
reason:         sleep killed by SIGSEGV
time:           Mon 25 May 2020 04:36:26 PM CEST
cmdline:        sleep 10m
package:        coreutils-8.22-24.el7
uid:            1000 (djuarezg)
count:          2
Directory:      /var/spool/abrt/ccpp-2020-05-25-16:36:26-27533
Reported:       https://abrtanalytics.web.cern.ch/faf/reports/bthash/5bb483b5ca0ea992b911f4d845479a085eab0f24
                https://retrace.fedoraproject.org/faf/reports/bthash/058896b486386a0060459830992f69a275e7f3fd
```

This seems to make that command useless when deploying private instances.

How is such hash specifically determined? Cannot seem to find it for now.

EDIT: https://github.com/abrt/faf/commit/cd8d687cdfaff3bc86d4a8d576d8d5175aaefaa6#diff-f99db2d089007acb7880d0cb22908f75R47 this is what checks the corresponding hash to see if there is a match.

djuarezg:
[...]

PR #942, opened 2020-09-17
pyfaf: problemtypes: core: Don’t ASCII-encode file names

ernestask:
Currently, when building a list of symbol strings, the file name ends up
being converted to ASCII and added to a Unicode string, which somehow
results in the byte-literal prefix being stored verbatim, which further
results in inconsistent hashes for identical crash threads.

This commit drops the encoding bit.

Closes https://github.com/abrt/faf/issues/922

### DIFF
--- a/src/pyfaf/problemtypes/core.py
+++ b/src/pyfaf/problemtypes/core.py
@@ -307,9 +307,7 @@ class CoredumpProblem(ProblemType):
             if i >= self.hashframes:
                 break
 
-            hashbase.append("{0} @ {1}".format(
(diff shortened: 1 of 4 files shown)


ernestask:
Would be nice if someone verified this. I haven’t really tested with more-exotic file names, but, looking at other problem plugins (Python), this is how it appears to be done.

ernestask:
Or I’ll just look into writing a test for this.

mgrabovsky (approved):
Looking great.

PR #946, opened 2020-09-23
addcompathashes: Remove import of hashlib

mkutlak:
Commit [0] removed only usage of hashlib and pylint is now complaining about
unused import.

[0] https://github.com/abrt/faf/commit/75f0d64a8214d3f303434fb8dc3c1cfa0c9d0daf


mgrabovsky:
Good catch. Thanks.

mgrabovsky (approved):
Good catch. Thanks.

Issue #948, opened 2020-10-10
Make newly released RPMs and containers available


mkutlak:
New ABRT Analytics has been released X days ago but RPMs are not available in [COPR](https://copr.fedorainfracloud.org/coprs/g/abrt/faf-el8/) nor is the container in [Dockerhub](https://hub.docker.com/r/abrt/faf/tags).

michalfabik:
Hi, I'm actually looking into this right now and can't figure out how to set up the quay.io build triggers. When trying to run a trigger by hand, I'm getting "Tag does not exist in remote git repo: 2.3.0 You must tag, then `git push --follow-tags`" from `tito`, even though the tag is in the master branch.

mkutlak:
```bash
# Stage 1: Build RPMs

FROM fedora:32
USER root
RUN dnf --assumeyes install git-core make tito
COPY . /src
WORKDIR /src
RUN \
    dnf --assumeyes builddep --spec faf.spec && \
    useradd --no-create-home builder && \
    chown --recursive --quiet builder. .
USER builder
RUN tito build --rpm
```

Yeah, the Dockfile assumes that your `origin` is pointing to `abrt/faf` and is `HTTPS` clone.

If any of that isn't true `tito` can't check the remote repository and outputs that error message.

The workaround:
Change your remote `origin` to `https://github.com/abrt/faf.git`.

``` bash
$ git remote set-url origin https://github.com/abrt/faf.git
```

mgrabovsky:
@michalfabik What is the status here? It appears that 2.3.1 has been pushed to Docker Hub but `latest` still points to 2.2.0.

mgrabovsky:
This is fixed now. I've just pushed the the [tags 2.4.0, 2.4 and _latest_](https://hub.docker.com/r/abrt/faf/tags?page=1&ordering=last_updated) for the latest release.

PR #949, opened 2020-10-10
dockerfile: Fix build stage

mkutlak:
With `COPY . /src` in the Dockerfile it assumed that `origin`
was pointing to `https://github.com/abrt/faf.git`.

If that wasn't true (like in my case, origin=git@github.com:abrt/faf.git)
`tito build` didn't have permissions (ssh key) to check the remote repo
and failed with `Tag does not exist in remote git repo`.

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -2,14 +2,13 @@
 
 FROM fedora:32
 USER root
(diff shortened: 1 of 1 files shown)


mkutlak:
Relevant: #948

michalfabik (approved):

PR #955, opened 2020-10-22
Fix LOB saving issue

### LINKED ISSUE #954: Persistent error when saving LOBs
The following error has been occurring on retrace.fp.org for a while now:
```
ERROR:root: Data must be either a bytestring or a file-like object
```

It's an error message from the method [`GenericTableBase.save_lob()`](https://github.com/abrt/faf/blob/d5d31c1/src/pyfaf/storage/generic_table.py#L99-L117). It is raised when the data passed to the function don't have the correct “shape”:
```python
if not isinstance(data, bytes) and not hasattr(data, "read"):
    raise FafError("Data must be either a bytestring or a file-like object")
```

mgrabovsky:
- Encode the JSON of an invalid report to a bytestring before saving it as a LOB.
- Add typing annotations to all methods of `GenericTableBase`.
- Add casts where necessary.
- Improve an error message in `save_lob()` to report the name of the LOB.
- Fix minor issues reported by Pylint.

Resolves #954

### DIFF
--- a/src/pyfaf/storage/generic_table.py
+++ b/src/pyfaf/storage/generic_table.py
@@ -1,7 +1,6 @@
 import errno
 import os
-
-from io import BufferedReader
+from typing import cast, BinaryIO, Optional, Union
 
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


xsuchy:
LGTM, but we should stop building retrace-server for f31 because of that

     ModuleNotFoundError: No module named 'werkzeug.middleware'

xsuchy (approved):

PR #957, opened 2020-10-24
generic_table: Remove redundant import

michalfabik:
See commit.


mgrabovsky (approved):
Good catch.

PR #961, opened 2020-11-04
webfaf: Add ad-hoc banner

michalfabik:
Add a quick and dirty way of putting up a banner notifying all visiors
of scheduled maintenance and similar issues.

While I was adding the possible discontinuation notice to fafreport, I thought I migh as well make it general purpose for similar occasions.

### DIFF
--- a/src/webfaf/config.py
+++ b/src/webfaf/config.py
@@ -28,6 +28,7 @@ class Config(object):
     MAIL_FROM = config.get("mail.from", 'no-reply@' + MAIL_SERVER)
     BRAND_TITLE = config.get("hub.brand_title", "ABRT")
     BRAND_SUBTITLE = config.get("hub.brand_subtitle", "Analytics")
+    BANNER = config.get("hub.banner", "")
     CACHE_TYPE = config.get("cache.type", "simple")
     MEMCACHED_HOST = config.get("cache.memcached_host", None)
     MEMCACHED_PORT = config.get("cache.memcached_port", None)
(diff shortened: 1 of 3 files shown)


src/webfaf/templates/base.html
```diff
@@ -30,6 +30,12 @@
     {% endblock js %}
   </head>
   <body>
+      <!--Banner notifiying of scheduled downtime etc. Uncomment the below div and edit text as needed.-->
+      <!--
+      <div class="alert alert-danger">
+              <b>Attention all users, here be an important message.</b>
+      </div>
+      -->
```
xsuchy, line 38: But... this will be shown to everyone. If you want to do that in upstream, can you wrap it in:

{%- if config['BANNER'] -%}

or you can steal the code from 

https://pagure.io/copr/copr/c/5ab2579adc40e39e056086babed5c1a608417666?branch=master

src/webfaf/config.py
```diff
@@ -28,6 +28,7 @@ class Config(object):
     MAIL_FROM = config.get("mail.from", 'no-reply@' + MAIL_SERVER)
     BRAND_TITLE = config.get("hub.brand_title", "ABRT")
     BRAND_SUBTITLE = config.get("hub.brand_subtitle", "Analytics")
+    BANNER = config.get("hub.banner")
```
xsuchy, line 31: Can you set the default? Either None or an empty string.

xsuchy (approved):

PR #962, opened 2020-11-22
storage: Use file extensions with package LOBs

michalfabik:
See commit.

### DIFF
--- a/src/pyfaf/storage/generic_table.py
+++ b/src/pyfaf/storage/generic_table.py
@@ -31,9 +31,8 @@ class GenericTableBase(object):
             raise FafError("'{0}' does not allow a lob named '{1}'".format(classname, name))
 
         pkstr = self.pkstr()
-        pkstr_long = pkstr
-        while len(pkstr_long) < 5:
-            pkstr_long = "{0}{1}".format("".join(["0" for i in range(5 - len(pkstr_long))]), pkstr_long)
+        pkstr_long = pkstr.rjust(5, '0')
+        pkstr = "{0}{1}".format(pkstr, self.get_lob_extension())
 
         lobdir = os.path.join(config["storage.lobdir"], classname, name,
                               pkstr_long[0:2], pkstr_long[2:4])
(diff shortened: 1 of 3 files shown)


src/pyfaf/storage/generic_table.py
```diff
@@ -43,6 +43,11 @@ def get_lob_path(self, name: str) -> str:
             if ex.errno != errno.EEXIST:
                 raise
 
+        for col in self.__table__._columns: #pylint: disable=no-member, protected-access
```
xsuchy, line 46: I do not like this for-loop (because of performance). Can we instead have GenericsTableBase.get_lob_extension() which will return empty string. And in `Package` it will be redefined to return extension based on `pkgtype` attribute?

src/pyfaf/storage/generic_table.py
```diff
@@ -32,6 +32,8 @@ def get_lob_path(self, name: str) -> str:
 
         pkstr = self.pkstr()
         pkstr_long = pkstr
+        pkstr = "{0}{1}".format(pkstr, self.get_lob_extension())
+
         while len(pkstr_long) < 5:
             pkstr_long = "{0}{1}".format("".join(["0" for i in range(5 - len(pkstr_long))]), pkstr_long)
```
mgrabovsky, line 38: While we are at it, this monstrosity can be replace with just
```python
pkstr_long = pkstr_long.rjust(5, '0')
```
if I'm reading it correctly.

michalfabik: Nice, thanks! Rewritten.

src/pyfaf/storage/migrations/versions/e1e54ec3137d_use_rpm_extension_with_lobs.py
```diff
@@ -0,0 +1,76 @@
...
+                 package.c.build_id,
+                 package.c.pkgtype]
+            )):
+        db_package = Package()
+        db_package.id = pkgid
+        db_package.build_id = build
+        db_package.pkgtype = pkgtype
+        ext_len = len(pkgtype) + 1
+        lobpath = db_package.get_lob_path("package")
+        if lobpath.endswith(".{}".format(pkgtype)) and os.path.isfile(lobpath[:-ext_len]):
```
xsuchy, line 58: This will technically work. But can you use instead of `pkgtype` rather  `db_package.get_lob_extension`. I believe it would be more semantically consistent. Hmm... but you know... this is one shot upgrade script. It is fine. :)

xsuchy (approved):

Issue #908, opened 2020-05-07
Container instructions do not work


djuarezg:
Regarding the instructions on https://hub.docker.com/r/abrt/faf-image, all wen well except for some minor adaptations I had to use:

```bash
podman volume create volume_name
podman pod create -p 5432:5432 -p 6379:6379 -p 8080:8080 --name pod_name
podman run --pod pod_name -v /tmp/podman_tests:/var/lib/pgsql/data --privileged -e POSTGRESQL_ADMIN_PASSWORD=scrt --name db -dit abrt/postgres-semver
podman exec db sh -c "psql -c \"SELECT 1 FROM pg_roles WHERE rolname='faf'\" | grep -q 1 || yes scrt | createuser -Ps faf"
podman run --pod pod_name --privileged --name faf-redis --hostname faf-redis -dit redis
podman run --pod pod_name --privileged --name faf -dit -e PGHOST=localhost -e PGUSER=faf -e PGPASSWORD=scrt -e PGPORT=5432 -e PGDATABASE=faf \
-e RDSBROKER=redis://faf-redis:6379/0 -e RDSBACKEND=redis://faf-redis:6379/0 abrt/faf-image:2.2.0
```

The problem is that with such instructions, when checking the webUI, I get:

```
Internal Server Error. This error has been logged and we will take care of it soon. Sorry. 
```

On the server side:

```
[2020-05-07 12:27:15] INFO:faf.PullReleases: Downloading releases from remote database
[2020-05-07 12:27:20] INFO:faf.PullReleases: Adding release 'rawhide' (ACTIVE)
[2020-05-07 12:27:20] INFO:faf.PullReleases: Adding release 'ln' (ACTIVE)
[2020-05-07 12:27:20] INFO:faf.PullReleases: Adding release 'l' (ACTIVE)
[2020-05-07 12:27:20] INFO:faf.PullReleases: Adding release '30' (ACTIVE)
[2020-05-07 12:27:20] INFO:faf.PullReleases: Adding release '31' (ACTIVE)
[2020-05-07 12:27:20] INFO:faf.PullReleases: Adding release '32' (ACTIVE)
[2020-05-07 12:27:20] INFO:faf.PullReleases: Adding release '33' (ACTIVE)
[uWSGI] getting INI configuration from /etc/uwsgi.ini
setting capability setgid [6]
setting capability setuid [7]
[...]

PR #963, opened 2020-11-25
container: Fix some podman commands in container readme

mzidek-gh:
Some commands used old names of containers.

Related: https://github.com/abrt/faf/issues/908


mgrabovsky (approved):
Looks good.

PR #967, opened 2021-01-12
Fix RHEL 8 builds

michalfabik:
See commits.

Signed-off-by: Michal Fabik <mfabik@redhat.com>

PR #968, opened 2021-02-13
celery_tasks: Remove future-dependent annotation

michalfabik:
Fallout from https://github.com/abrt/faf/pull/967. Type annotation
without futures causes a crash when accessing the FAF web interface
because the class in the annotation is only defined later in the file.
Moving the code around causes the same issue with a different class so
I'm just dropping the annotation.


mzidek-gh (approved):

mzidek-gh:
ACK.

PR #972, opened 2021-03-17
Fix issues detected by Pylint

mgrabovsky:
This fixes some long-standing errors and warnings from Pylint so that we can marvel at the green check mark again :heavy_check_mark:

* Re-raise exceptions using `raise … from …`.
* Call `super()` with no arguments in constructors.
* Make `__str__()` return a string in all cases.
* Suppress warnings about wrong import order.
* Remove some unused methods.
* Wrap long lines.


mgrabovsky:
Ready for review now.

michalfabik (approved):

michalfabik:
LGTM

PR #973, opened 2021-03-18
WIP: storage: Drop autocommit parameter

mgrabovsky:
The autocommit mode [will be deprecated](https://docs.sqlalchemy.org/en/13/orm/session_transaction.html#session-autocommit) in SQLAlchemy 1.4. Although 1.4 has only been released in March 2021 and 1.3 is still going strong as of this month, we should begin to prepare for a migration to 1.4 (or even 2.0) in the future.

Related: #974


mgrabovsky:
This seems like a lot of work for little gain at the moment. Closing this for now.

PR #981, opened 2021-06-24
Only accept Fedora GA releases

mgrabovsky:
When retrieving the list of active Fedora releases from Fedora PDC, only consider GA (general availability) releases, i.e. ignore `*-updates`, `*-updates-testing` etc.

This should prevent EOL'd release from reappearing on the summary page, since `faf pull-releases` used to switch their state to `ACTIVE` back from `EOL` due to the confusion in release types.

### DIFF
--- a/src/pyfaf/opsys/centos.py
+++ b/src/pyfaf/opsys/centos.py
@@ -227,7 +227,7 @@ class CentOS(System):
                 .filter(Build.release.like("%%.el%%"))
                 .all())
 
-    def check_pkgname_match(self, packages, parser) -> None:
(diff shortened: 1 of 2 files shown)


msrb (approved):
LGTM +1

In future, we should probably move away from PDC as (AFAIK) it is no longer maintained. Bodhi should be the authoritative source for this kind of information.

mgrabovsky:
> LGTM +1
> 
> In future, we should probably move away from PDC as (AFAIK) it is no longer maintained. Bodhi should be the authoritative source for this kind of information.

Ah, good to know. I'll file an issue for that. Thanks for the review.

PR #996, opened 2021-11-26
Sonar linting

michalfabik:
I'm experimenting with SonarQube so here are a few fixes it's suggested - mostly needlessly nested conditions and some minor style issues.  
  
Also, one actual bug in a table definition in `storage`.


mgrabovsky (approved):
Looks great.

PR #999, opened 2022-01-07
Require Celery >= 5.2.2

mgrabovsky:
Bump Celery dependency due to the [CVE-2021-23727][1] vulnerability which affects Celery versions prior to 5.2.2.

[1]: https://github.com/advisories/GHSA-q4xr-rc97-m4xx

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Changed
+- Bumped Celery dependency to >= 5.2.2 due to a security bug in earlier versions
 
 ## [2.5.0] - 2021-10-18
 ### Changed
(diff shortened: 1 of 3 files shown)


mgrabovsky:
Both Celery 5.2.3 and WTForms-SQLAlchemy are now built in our Copr, so I suppose we can proceed with the upgrade.

WTForms-SQLAlchemy is needed separately because it was split out from WTForms in 3.x and it's the reason why Packit Fedora 36+ builds keep failing. Down the road, the builds will keep failing anyway because WTForms-SQLAlchemy is not packaged in Fedora. (I filed a [review request](https://bugzilla.redhat.com/bugzilla/show_bug.cgi?id=2055622), but I'm now thinking that it's unnecessary and we might keep it in the Copr only.)  That leads me to considering dropping the Packit integration since it will become essentially useless.

I'm merging this for now and will create a separate PR for the WTForms-SQLAlchemy migration.

PR #1000, opened 2022-01-11
Fix build on Fedora 34+

mgrabovsky:
* The `dbname` parameter has been changed to `database` in python-testing.sqlalchemy.
* Bump dependency on python-testing.postgresql in spec file.

### DIFF
--- a/faf.spec
+++ b/faf.spec
@@ -40,7 +40,7 @@ BuildRequires: python3-alembic
 BuildRequires: python3-setuptools
 BuildRequires: python3-bugzilla >= 2.0
 BuildRequires: python3-psycopg2
-BuildRequires: python3-testing.postgresql < 1.2
+BuildRequires: python3-testing.postgresql >= 1.3.0
 BuildRequires: python3-rpm
 BuildRequires: python3-sqlalchemy
 BuildRequires: python3-koji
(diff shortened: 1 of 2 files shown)


mgrabovsky:
Well, this fixes Fedora 35 for now.

F34 stays broken because the [pertaining update](https://bodhi.fedoraproject.org/updates/FEDORA-2022-7a679b2d28) is still in testing.

Rawhide/F36, on the other hand, is now broken because [WTForms](https://src.fedoraproject.org/rpms/python-wtforms) has been bumped to 3.0.x, dropping all modules in `wtforms.ext`. We only rely on `wtforms.ext.sqlalchemy` which is now a [separate library](pypi.org/project/WTForms-Alchemy/) so it might need packaging.

mgrabovsky:
I'll merge this now as it fixes at least a part of the problem.

WTForms-SQLAlchemy is now built [in Copr](https://copr.fedorainfracloud.org/coprs/mgrabovs/python-wtforms-sqlalchemy/) for Fedoras and EPEL 8, so it can be practically used. I've also created a [package review request](https://bugzilla.redhat.com/bugzilla/show_bug.cgi?id=2055622) so it might land in official repos one day as well.

Issue #461, opened 2015-09-09
More user friendly error messages


jfilak:
FAF replies with the following error message if OS is not supported:

```
Server responded with an error: 'uReport data is invalid.'
```

kshitija08:
@marusak I would like to work on this issue. Kindly, guide me further.

marusak:
Hi @kshitija08 
happy to hear you are interested :) 

When you report into FAF, the ureport may be refused and you only get response from server `uReport data is invalid.` However server logs why the ureport was invalid. It would be great, if server would inform more why the ureport was refused.

You may try this easily. Create some problem and report it. Then in dumpdir (problem directory stored in `/var/spool/abrt/...` change for example in file `os_info` `ID=fedora` to `id=rhel`. Then report it again and you get `Server responded with an error: 'uReport data is invalid.'`. It would be much better if you would get something like `Server responded with an error: 'OS 'rhel' is not supported by server.'`

ArnaudKleinveld:
Would love to see progress on this issue. Since the installation of F34 I have been plagued with java 11 crashes and trying to report it using the abrt gui it's returning a 'uReport data is invalid' and 'report_uReport' exited with 1'. Anything I can do to help?

PR #1002, opened 2022-02-17
More informative error message on invalid uReport

michalfabik:
Return actual exception to reporter when an uReport is rejected. So far,
this was only done in one specific situation and otherwise an overly
generic error message was returned.

Resolves: https://github.com/abrt/faf/issues/461


mgrabovsky (approved):
Looks good to me. Hopefully the errors from lower levels are more informative.

PR #1005, opened 2022-03-09
Migrate to WTForms-SQLAlchemy

mgrabovsky:


mgrabovsky:
[Builds and tests fine](https://copr.fedorainfracloud.org/coprs/g/abrt/faf-el8-devel/build/3652147/) on Fedora 35+ and EPEL 8. Not sure if we should care about F34 at this point (EOL in [mid-May](https://fedorapeople.org/groups/schedule/f-36/f-36-key-tasks.html)).

mgrabovsky:
[Builds (still) go through](https://copr.fedorainfracloud.org/coprs/g/abrt/faf-el8-devel/build/4190861/) on Fedora 35+ and EPEL 8, so I'm merging this. We'll have Fedora < 34 unsupported as of now.

Issue #1008, opened 2022-04-25
Remove the “Report a bug” link from the footer


mgrabovsky:
We keep getting bug reports, albeit infrequently, for crashed applications in our bug tracker. It appears that people interpret the “Report a bug” link the footer of the web UI as a way to report a bug in the pertaining application rather in ABRT Analytics itself.

One easy solution is to remove the link completely or move it somewhere else. Perhaps a careful rephrasing might help as well?

mgrabovsky:
Fixed in #1010.

PR #1010, opened 2022-04-26
Deletes "Report a Bug" option

pmbechard:
Solves issue #1008 Remove the “Report a bug” link from the footer


mgrabovsky (approved):
Thanks for the PR.

PR #1012, opened 2022-07-26
Update README for containers

msrb:


container/README.md
```diff
@@ -6,29 +6,29 @@
...
 A Redis container is required to run and schedule faf actions from the web UI (see below).
 Communication between ABRT Analytics, the database and Redis is ensured by running all three containers
 in the same podman pod.
 
-`podman volume create <volume_name>`  
-`podman pod create -p 5432:5432 -p 6379:6379 -p 8080:8080 --name <pod_name>`  
-`podman run --pod <pod_name> -v <volume_name>:/var/lib/pgsql/data -e POSTGRESQL_ADMIN_PASSWORD=scrt --name db -dit abrt/postgres-semver`
+`podman volume create faf-volume`
+`podman pod create -p 5432:5432 -p 6379:6379 -p 8080:8080 --name faf-pod`
+`podman run --pod faf-pod -v faf-volume:/var/lib/pgsql/data -e POSTGRESQL_ADMIN_PASSWORD=scrt --name db -dit quay.io/abrt/faf-db`
```
mgrabovsky, line 17: Can you please convert all these blocks to the indented form?
```suggestion
    podman volume create faf-volume
    podman pod create -p 5432:5432 -p 6379:6379 -p 8080:8080 --name faf-pod
    podman run --pod faf-pod -v faf-volume:/var/lib/pgsql/data -e POSTGRESQL_ADMIN_PASSWORD=scrt --name db -dit quay.io/abrt/faf-db
```

mgrabovsky (changes requested):
Looks OK, just one small suggestion while we're at it.

By the way, since when do we publish on quay.io?

container/README.md
msrb: Good point -- updated.

msrb:
> By the way, since when do we publish on quay.io?

That is a good question and I am trying to figure it out :) The namespace already exists, and the images were probably pushed to both docker.io and quay.io by some automation... 9 months ago. But not since then.

mgrabovsky (approved):

PR #1013, opened 2022-08-09
Modernize containers

msrb:


container/Dockerfile
```diff
@@ -1,21 +1,23 @@
 # Stage 1: Build RPMs
 
-FROM fedora:32
+FROM fedora:36
 USER root
+RUN dnf --assumeyes install 'dnf-command(copr)' && dnf --assumeyes copr enable mgrabovs/python-wtforms-sqlalchemy
```
mgrabovsky, line 5: I think I'd prefer to move the repo under the abrt Copr org/group.

mgrabovsky: Actually, I already have a [package review request](https://bugzilla.redhat.com/show_bug.cgi?id=2055622) open that I forgot about.

mgrabovsky: I'm actually going to build the package in the [faf-el8 Copr](https://copr.fedorainfracloud.org/coprs/g/abrt/faf-el8/builds/) so that we can pull it here.

martinky82:
I'll need to use more recent Fedora as a base due to dependency on python3-bugzilla 3.2.0 which is not available on Fedora older than 35 (and there it is built against python 3.10).

container/Dockerfile
mgrabovsky: @msrb It's built now, can you please update the PR?

mgrabovsky: Hm, it just occurred to me that we might need to build a couple more packages for all this to work correctly on RHEL 8, but I don't have the capacity to look into it right now.

msrb: @mgrabovsky sorry for the delay here.

I will try to update things here by the end of this week.

mgrabovsky:
@msrb Could you please try to update this? There's some demand for newer container images, but the build seems to be broken at the moment and it seems these improvement would be nice to have there.

mgrabovsky:
Actually, I'll just open a new PR if that's fine with you.

PR #1015, opened 2022-09-20
Remove SMTP logging handler

mgrabovsky:
Sending alerts for exception should not be a responsibility of the application but a monitoring system. Moreover, the SMTP handler is not configured correctly in production and only clutters log files with exceptions.


msrb (approved):
+1
I agree, that functionality doesn't belong in here.

michalfabik (approved):
Looks good to me.

PR #1022, opened 2022-10-24
Fix GitHub build & test workflow

mgrabovsky:


msrb (approved):
lgtm ;)

.github/workflows/check.yml
```diff
@@ -7,26 +7,26 @@ on:
...
+      image: quay.io/fedora/fedora:latest
     runs-on: ubuntu-latest
     steps:
       - name: Check out sources
         uses: actions/checkout@v3
 
       - name: Install build and check environment
         run: |
-          dnf --assumeyes install --setopt=install_weak_deps=False \
+          dnf install --assumeyes --setopt=install_weak_deps=False \
```
michalfabik, line 18: Oh no, I'm too late! Anyway, FWIW, I think the former was "more correct" because `--asumeyes` is an argument of `dnf` itself, not the `install` command. I actually saw a PR the other day that was fixing this specifically, though I don't remember why/if it mattered.

PR #1026, opened 2023-01-10
opsys: Allow carets in package version

### LINKED ISSUE #484: view abtr reports per machine
Having occurrences of reports counted once per user is for developers better statistic which could better priories bug reports. I would like to see special view from web UI for this feature.

In Fedora all traffic should be anonymous so some local database with already reported crashes + sending a flag(whether it was already reported by this user or not) to the server can workaround mentioned issue. RFE was discussed and pre-approved by The Code Master (@jfilak)

mgrabovsky:
The [Fedora packaging guidelines][1] require a caret to be used in the version string for package snaphots.

[1]: https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/#_snapshots

Fixes #484

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Fixed
+- Allow the caret (`^`) in Fedora and CentOS package versions. Fedora requires it [for snapshots](https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/#_snapshots)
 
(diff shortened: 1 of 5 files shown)


mgrabovsky:
@msrb Can you please review this? The test goes through on my machine (once combined with #1022).

msrb:
Amazing ;) +1 for the caret change.

I am a bit puzzled by the `tests/faftests/__init__.py.in` file though. What's the purpose (?) Or was it added by accident?

mgrabovsky:
Oh, snap, good catch! That belongs to the other PR. That's what code review is for :smile:

msrb (approved):

PR #1031, opened 2024-04-06
Improve an error message

jvoisin:
Knowing what erroneous value was passed will make debugging easier.


TitouanT:
I thin the second format index should be updated to `{1}` instead of the original `{0}`, the current change would display the offending value twice

jvoisin:
Oops, good catch.

msrb (approved):

msrb:
Thank you ;)

